PR #21, opened 2021-02-06
Several misc. fixes, cleanups, features

ericaltendorf:
- Fix logic for identifying plotting processes
- Update sample config file to be valid and better commented
- Add a job-progress histogram to the interactive display
- Various minor fixes and cleanups

### DIFF
--- a/interactive.py
+++ b/interactive.py
@@ -11,12 +11,6 @@ import archive
 import manager
 import reporting
 
-def window_width(window):
-    return window.getmaxyx()[1]
-
-def window_height(window):
-    return window.getmaxyx()[0]
-
 class Log:
     entries = []
     cur_pos = 0
@@ -81,7 +75,7 @@ def curses_main(stdscr):
 
     # Page layout.  Currently requires at least ~40 rows.
     # TODO: make everything dynamically resize to best use available space
-    header_height = 2
+    header_height = 3
     jobs_height = 10
     dirs_height = 14
     logscreen_height = n_rows - (header_height + jobs_height + dirs_height)
@@ -97,7 +91,7 @@ def curses_main(stdscr):
     refresh_period = int(sched_cfg['polling_time_s'])
 
     stdscr.nodelay(True)  # make getch() non-blocking
-    stdscr.timeout(5000)   # this doesn't seem to do anything....
+    stdscr.timeout(2000)
 
     header_win = curses.newwin(header_height, n_cols, header_pos, 0)
     log_win = curses.newwin(logscreen_height, n_cols, logscreen_pos, 0)
@@ -111,18 +105,24 @@ def curses_main(stdscr):
 
     while True:
 
-        # todo: none of this resizing works
-        (n_rows, n_cols) = map(int, stdscr.getmaxyx())
+        # TODO: handle resizing.  Need to (1) figure out how to reliably get
+        # the terminal size -- the recommended method doesn't seem to work
+        #    (n_rows, n_cols) = map(int, stdscr.getmaxyx()) ...doesn't work
+        #    ...map(int, os.popen('stty size', 'r').read().split() ...may work
+        # and then (2) implement the logic to resize all the subwindows as above
+
         stdscr.clear()
         linecap = n_cols - 1
         logscreen_height = n_rows - (header_height + jobs_height + dirs_height)
 
         elapsed = (datetime.datetime.now() - last_refresh).total_seconds() 
+        full_refresh = False
         if (elapsed < refresh_period):
             # Lightweight; does virtually no work if there are no new jobs.
             jobs = Job.get_running_jobs_w_cache(dir_cfg['log'], jobs)
 
         else:
+            full_refresh = True
             # Full refresh
             last_refresh = datetime.datetime.now()
             jobs = Job.get_running_jobs(dir_cfg['log'])
@@ -134,7 +134,7 @@ def curses_main(stdscr):
                 (started, msg) = manager.maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg)
                 if (started):
                     log.log(msg)
-                    plotting_status = '<just started plot>'
+                    plotting_status = '<just started job>'
                     jobs = Job.get_running_jobs_w_cache(dir_cfg['log'], jobs)
                 else:
                     plotting_status = msg
@@ -162,14 +162,11 @@ def curses_main(stdscr):
         dst_prefix = os.path.commonpath(dir_cfg['dst'])
         arch_prefix = dir_cfg['archive']['rsyncd_path']
 
-        # Render
-        stdscr.clear()
-
         # Header
         header_win.addnstr(0, 0, 'Plotman', linecap, curses.A_BOLD)
-        header_win.addnstr(' %s (refresh %ds/%ds)' %
-                (datetime.datetime.now().strftime("%H:%M:%S"), elapsed, refresh_period),
-                linecap)
+        timestamp = datetime.datetime.now().strftime("%H:%M:%S")
+        refresh_msg = "now" if full_refresh else ('%ds/%ds' % (elapsed, refresh_period))
+        header_win.addnstr(' %s (refresh %s)' % (timestamp, refresh_msg), linecap)
         header_win.addnstr('  |  <P>lotting: ', linecap, curses.A_BOLD)
         header_win.addnstr(
                 plotting_status_msg(plotting_active, plotting_status), linecap)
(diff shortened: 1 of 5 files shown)


reporting.py
```diff
@@ -13,9 +14,53 @@ def abbr_path(path, putative_prefix):
...
+    elif n == 1:
+        return '.'
+    elif n == 2:
+        return ':'
+    elif n == 3:
+        return ';'
+    elif n >= 4:
+        return '!'
+    else:
+        return 'X'  # Should never be negative
```
altendky, line 45:
```suggestion
n_to_char_mapping = dict(enumerate(" .:;"))

def n_to_char(n):
    if n < 0:
        return 'X'  # Should never be negative

    return n_to_char_mapping.get(n, "!")
```
Might also like mapping to braille dots.  I feel a little inappropriate misusing those characters, but they do work nicely for 'modern' spinners and such.

reporting_test.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python3
+
+from unittest.mock import patch
```
altendky, line 3: I very rarely mock.

reporting_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+import os
+import pyfakefs
+import unittest
```
altendky, line 7: I always use pytest.

reporting.py
```diff
@@ -13,9 +14,53 @@ def abbr_path(path, putative_prefix):
...
-    return ', '.join(['%d:%d' % ph_subph for ph_subph in phases])
+    if not max_num or len(phases) <= max_num:
+        return ' '.join(['%d:%d' % pair for pair in phases])
+    else:
+        n_first = math.floor(max_num / 2)
+        n_last = max_num - n_first
+        n_elided = len(phases) - (n_first + n_last)
+        return (' '.join(['%d:%d' % pair for pair in phases[:n_first]]) +
+            " [+%d] " % n_elided +
+            ' '.join(['%d:%d' % pair for pair in phases[n_first + n_elided:]]))
```
altendky, line 27: More variables, less one-liner here.  `:]`

altendky (changes requested):
Sorry, but I didn't get to a whole lot of review of correctness.  Something I need to work on and also something that follows from 1) me being new to this code and 2) us being new to each other's 'style' preferences so that stuff jumps out etc.  Hopefully the explanations are either things you don't know or aspects you hadn't considered.

Compound PRs addressing multiple things generally end up being way harder to review since the reviewer has to dissect which bit goes with what.  Also, it means you are pretty likely to have several things held up from getting merged over some discussion related to just one little bit of the PR.  

```python
class Log:
    entries = []
    cur_pos = 0
```

I personally use https://attrs.org for pretty much all of my classes.  But, to this approach...  Those aren't defining attributes of the class instances, but attributes of the class itself.  Effectively globals that are just accessed via the class.  If you `self.cur_pos = 3` then you are adding an instance attribute shadowing the class attribute.  If you `self.entries.append(3)` you are mutating the single class attribute, not an instance specific list.  So, all instances will see that new value on the end of the list.  This is related to the common mutable function parameter default confusion.

FYI about GitHub reviews and suggestions.  If you like a suggestion exactly as is you can either commit it individually or group it together with others and commit them in a batch by clicking the corresponding buttons.  It can be convenient for you and, in cases where you care, it lists the suggester as an author on the commit as well.  (I don't care for my contributions but would for other people's)

image


You've got tests now.  Add CI next.  `:]`

altendky:
If you want to be picky about your commit messages...  GitHub suggests a length limit of 50.

image

If you want focused granular commits, `also` and `and` etc are hints you aren't achieving the goal.

As far as granular PRs...  at various points I've made compound PRs because there's a lot of stuff mixing together.  When I'm being good, I go back and break off auxiliary PRs.  As each of the pieces gets merged and you catch up the original branch, it'll get more and more focused.

interactive.py
```diff
@@ -111,18 +105,24 @@ def curses_main(stdscr):
...
     while True:
 
-        # todo: none of this resizing works
-        (n_rows, n_cols) = map(int, stdscr.getmaxyx())
+        # TODO: handle resizing.  Need to (1) figure out how to reliably get
+        # the terminal size -- the recommended method doesn't seem to work
+        #    (n_rows, n_cols) = map(int, stdscr.getmaxyx()) ...doesn't work
+        #    ...map(int, os.popen('stty size', 'r').read().split() ...may work
+        # and then (2) implement the logic to resize all the subwindows as above
+
```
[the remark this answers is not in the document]
ericaltendorf: Addressed here; will keep in mind in the future as well, thanks.

interactive.py
```diff
@@ -111,18 +105,24 @@ def curses_main(stdscr):
...
         stdscr.clear()
         linecap = n_cols - 1
         logscreen_height = n_rows - (header_height + jobs_height + dirs_height)
 
         elapsed = (datetime.datetime.now() - last_refresh).total_seconds() 
+        full_refresh = False
         if (elapsed < refresh_period):
```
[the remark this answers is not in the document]
ericaltendorf: This one is a little awkward because it's not recording that the refresh is needed, in order to do the refresh later, it's recording that we're doing the refresh right here and now, in order to display that fact later.  So `doing_full_refresh` or `full_refresh_done` might be more accurate.  Thoughts?

interactive.py
```diff
@@ -111,18 +105,24 @@ def curses_main(stdscr):
...
             # Lightweight; does virtually no work if there are no new jobs.
             jobs = Job.get_running_jobs_w_cache(dir_cfg['log'], jobs)
 
         else:
+            full_refresh = True
```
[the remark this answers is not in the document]
ericaltendorf: I think I need this line, see line 169 below,
```refresh_msg = "now" if full_refresh else ('%ds/%ds' % (elapsed, refresh_period))```

File: interactive.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (ericaltendorf) [reply interactive.py:120]: Makes sense.

File: interactive.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (ericaltendorf) [reply interactive.py:125]: ack.

File: interactive.py [hunk 8]
```diff
@@ -162,26 +162,27 @@ def curses_main(stdscr):
...
-        stdscr.clear()
-
         # Header
         header_win.addnstr(0, 0, 'Plotman', linecap, curses.A_BOLD)
-        header_win.addnstr(' %s (refresh %ds/%ds)' %
-                (datetime.datetime.now().strftime("%H:%M:%S"), elapsed, refresh_period),
-                linecap)
+        timestamp = datetime.datetime.now().strftime("%H:%M:%S")
+        refresh_msg = "now" if full_refresh else ('%ds/%ds' % (elapsed, refresh_period))
+        header_win.addnstr(' %s (refresh %s)' % (timestamp, refresh_msg), linecap)
```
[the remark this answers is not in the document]
ericaltendorf: Nice.

reporting.py
```diff
@@ -13,9 +14,53 @@ def abbr_path(path, putative_prefix):
...
+
+
+def n_at_ph(jobs, ph):
+    return len([j for j in jobs if j.progress() == ph])
```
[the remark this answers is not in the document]
ericaltendorf: That's clever.  I was looking for something like "count()" that would take a bool-returning function, kind of like map, but couldn't find one.  This works.

reporting_test.py
ericaltendorf: As in, you rarely need to, or there's another mechanism you prefer, or you structure your code so you don't need mock-style mechanisms?

reporting_test.py
ericaltendorf: Hmm, OK, added an issue to consider switching to pytest.

reporting.py
ericaltendorf: OK :)

reporting.py
ericaltendorf: Nice.  I took a slightly different approach to this to make the "!" treated more like the other chars.

Braille dots is an interesting option I hadn't thought of.  I did figure there would be some extended charset characters I could use but wasn't sure if I wanted to step out of ASCII.

ericaltendorf:
Thanks for all the comments.  Addressed most, will leave some larger ones (consider use of attrs, pytest) to later.

reporting_test.py
altendky: I try to avoid the need.  I see tests as a second use of your code.  If it's hard to test, it can mean your code isn't as flexible as it ought to be.  So yeah, different code structure often.  Lots of details to discuss in each actual case.

job.py
```diff
@@ -57,6 +57,13 @@ def get_running_jobs(logroot):
                     jobs.append(Job(proc, logroot))
         return jobs
 
+    def is_plotting_cmdline(cmdline):
+        return (len(cmdline) >= 4 and
+            'python' in cmdline[0] and
+            'venv/bin/chia' in cmdline[1] and
+            'plots' == cmdline[2] and
+            'create' == cmdline[3])
```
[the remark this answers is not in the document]
ericaltendorf: Nice, I'm learning a lot.

PR #27, opened 2021-02-12
fix: gracefully handle disappearing procs

ericaltendorf:
and fix inconsistencies between versions of job-getter funcs by combining them

### DIFF
--- a/job.py
+++ b/job.py
@@ -55,7 +55,7 @@ class Job:
     # These are dynamic, cached, and need to be udpated periodically
     phase = (None, None)   # Phase/subphase
 
-    def get_running_jobs(logroot, cached_jobs=None):
+    def get_running_jobs(logroot, cached_jobs=()):
         '''Return a list of running plot jobs.  If a cache of preexisting jobs is provided,
            reuse those previous jobs without updating their information.  Always look for
            new jobs not already in the cache.'''
--- a/job.py
+++ b/job.py
@@ -6,6 +6,7 @@ from enum import Enum, auto
 from subprocess import call
 import argparse
 
+import contextlib
 import logging
 import os
 import re
@@ -60,15 +61,14 @@ class Job:
            reuse those previous jobs without updating their information.  Always look for
            new jobs not already in the cache.'''
         jobs = []
-        if cached_jobs:
-            cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
+        cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
 
         for proc in psutil.process_iter(['pid', 'cmdline']):
             # Ignore processes which most likely have terminated between the time of
             # iteration and data access.
             with contextlib.suppress(psutil.NoSuchProcess):
                 if is_plotting_cmdline(proc.cmdline()):
-                    if cached_jobs and proc.pid in cached_jobs_by_pid.keys():
+                    if proc.pid in cached_jobs_by_pid.keys():
                         jobs.append(cached_jobs_by_pid[proc.pid])  # Copy from cache
                     else:
                         jobs.append(Job(proc, logroot))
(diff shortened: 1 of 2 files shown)


job.py
```diff
@@ -55,31 +55,26 @@ class Job:
...
+                if is_plotting_cmdline(proc.cmdline()):
+                    if cached_jobs and proc.pid in cached_jobs_by_pid.keys():
+                        jobs.append(cached_jobs_by_pid[proc.pid])  # Copy from cache
+                    else:
+                        jobs.append(Job(proc, logroot))
+            except psutil.NoSuchProcess:
+                # Generally this is a no-op; the process in question terminated
+                # after it was returned from the iter but before we got a
+                # chance to inspect its info.
+                pass
```
altendky, line 77: I don't know if it actually reads better here or not, mostly due to the comment, but fyi...

https://docs.python.org/3.9/library/contextlib.html#contextlib.suppress

```suggestion
            # Ignore processes which most likely have terminated between the time of
            # iteration and data access.
            with contextlib.suppress(psutil.NoSuchProcess):
                if is_plotting_cmdline(proc.cmdline()):
                    if cached_jobs and proc.pid in cached_jobs_by_pid.keys():
                        jobs.append(cached_jobs_by_pid[proc.pid])  # Copy from cache
                    else:
                        jobs.append(Job(proc, logroot))
```

job.py
```diff
@@ -55,31 +55,26 @@ class Job:
     # These are dynamic, cached, and need to be udpated periodically
     phase = (None, None)   # Phase/subphase
 
-    def get_running_jobs(logroot):
-        '''Return the list of running plot jobs, reading the process table from scratch.'''
+    def get_running_jobs(logroot, cached_jobs=None):
```
altendky, line 58: I think it could be a small bit tidier to just default this to an empty (immutable) collection.  Ought to drop the first `if` and the more complex if in the loop simplifies slightly.

```suggestion
    def get_running_jobs(logroot, cached_jobs=()):
```

altendky (approved):
Note that I'm not checking completeness of the refactor.  There may be leftover references to the removed function.  Normally tests with coverage checks would account for that (mostly).

Approving despite suggestions since the suggestions aren't obviously significantly better.  But maybe still of interest.

job.py
```diff
@@ -55,31 +55,23 @@ class Job:
...
-
-    def get_running_jobs_w_cache(logroot, existing_jobs):
-        '''Return the list of running plot jobs, returning previous jobs if they still exist, and
-           new jobs intialized.  Does not update info on existing jobs, which means information in
-           the list of jobs will be from different times.  Use this when calling frequent updates,
-           and use get_running_jobs() to force a full refresh.'''
-        jobs = []
-        existing_jobs_by_pid = { j.proc.pid: j for j in existing_jobs }
+        if cached_jobs:
+            cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
```
altendky, line 64:
```suggestion
        cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
```

job.py
```diff
@@ -55,31 +55,23 @@ class Job:
...
-            if is_plotting_cmdline(proc.cmdline()):
-                if proc.pid in existing_jobs_by_pid.keys():
-                    jobs.append(existing_jobs_by_pid[proc.pid])  # Copy from cache
-                else:
-                    jobs.append(Job(proc, logroot))
+            # Ignore processes which most likely have terminated between the time of
+            # iteration and data access.
+            with contextlib.suppress(psutil.NoSuchProcess):
+                if is_plotting_cmdline(proc.cmdline()):
+                    if cached_jobs and proc.pid in cached_jobs_by_pid.keys():
```
altendky, line 71:
```suggestion
                    if proc.pid in cached_jobs_by_pid.keys():
```

altendky (changes requested):
Sorry, the benefit of using an empty iterable as the default was to eliminate the special case handling code of when it was `None`.  Just makes it a tad easier to read.  Especially the second if where the `and` short-circuit was being leveraged to avoid accessing a not-defined variable.  (where I think both the maybe-defined variable and the short-circuit dependency are a bit smelly individually)

PR #32, opened 2021-02-18
fix: resolve all remote archive paths

fdachille:
Adding `-a` to the `df` command args shows all mount points.

During archive operation, remote filesystems are queried for size information. Some mount points are not shown by the `df -BK` command, but are shown by adding `-a` like `df -aBK`. Perhaps this happens [when a device is mounted to two different locations with a parent/child relationship](https://www.suse.com/support/kb/doc/?id=000019022). This change lists all the mounted directories (even pseudo ones) and fixes the problem of hidden mounts for me in the plotman archive operation.

### DIFF
--- a/archive.py
+++ b/archive.py
@@ -49,7 +49,7 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = { }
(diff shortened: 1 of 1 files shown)


ericaltendorf (approved):
Thank you!

PR #44, opened 2021-03-10
[WIP] only consider starting plots on directories that are present currently in config file

alfonsoperez:
Saw the issue reported in Keybase, just threw this in quick although haven't really thought through it yet


alfonsoperez:
that makes sense!, will close, and keep an eye on the dev branch for next time.

PR #45, opened 2021-03-14
Fix kill command so that it deletes all the temp files.

fdachille:
The get_temp_files function detects the log file open in twice so it was returning it twice, and the deletion of temporary files during the `kill` command was attempting to delete the log file twice which failed the second attempt. By returning all the temp files in a set, it only deletes each file once and succeeds in deleting all the temp files.

### DIFF
--- a/job.py
+++ b/job.py
@@ -289,10 +289,11 @@ class Job:
         self.proc.resume()
 
     def get_temp_files(self):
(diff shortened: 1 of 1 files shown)


ericaltendorf (approved):
Looks good, thanks!

PR #47, opened 2021-03-17
Development

fdachille:
Allow a different number of plots per tmp dir.

This sets up a pattern of overriding certain parameters per tmp file. In this case, the max number of jobs per tmp file defaults to the scheduler config 'tmpdir_max_jobs', but allows each tmp dir to override with it's own unique value for max jobs. This same pattern could be used to specialize other attributes per tmp dir.

The main use for this is for handling heterogeneous sizes of tmp drives with a single plotman configuration. For example, this would make it easy to support both 1TB, 2TB, and 6.4TB tmp dirs in the same configuration while fully utilizing the available space of each.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -19,6 +19,16 @@ directories:
                 - /mnt/tmp/02
                 - /mnt/tmp/03
 
+       # Optional: Allows overriding some characteristics of certain tmp
+       # directories. This contains a map of tmp directory names to
+       # attributes. If a tmp directory and attribute is not listed here,
+       # it uses the default attribute setting from the main configuration.
+       tmp_overrides:
+                # In this example, /mnt/tmp/00 is larger than the other tmp
+                # dirs and it can hold more plots than the default.
+                "/mnt/tmp/00":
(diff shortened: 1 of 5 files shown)


config.yaml
```diff
@@ -19,6 +19,16 @@ directories:
                 - /mnt/tmp/02
                 - /mnt/tmp/03
 
+       # Optional: Allows overriding some characteristics of certain tmp
+       # directories. This contains a map of tmp directory names to
+       # attributes. If a tmp directory and attribute is not listed here,
+       # it uses the default attribute setting from the main configuration.
```
ericaltendorf, line 25: Can we add a note here listing the attributes for which overrides are currently supported (right now, just `tmpdir_max_jobs`)?

ericaltendorf (changes requested):
Thanks so much for this, much cleaner than I was fearing it would end up being.  Just one request.

config.yaml
ericaltendorf: actually i'll just add this in a followup.

ericaltendorf (approved):

PR #51, opened 2021-03-22
Survive unexpected "plots create" invocation

serpent213:
Calling "chia plots create -h" caused a crash when its command line
was parsed.


altendky:
The bare `except: pass` hazards losing a lot of debug info for other errors.  We should probably make `Job` raise a particular exception for 'this case' (however it gets scoped) and handle just that with some sort of output somewhere.  Also, `-h`/`--help` should probably be explicitly supported and ignored without any exceptions involved, caught or otherwise.

altendky:
This should be fixed in https://github.com/ericaltendorf/plotman/pull/103.  If I've missed part of your problem, please do let me know.

Issue #38, opened 2021-03-04
Be robust to different platforms' psutil function availability


ericaltendorf:
psutil offers some functions which are platform dependent.  we could be more robust and simply not report ones that aren't available.

e.g., fix this stacktrace:
```File "/home/chia/plotman-main/interactive.py", line 207, in curses_main
    jobs_win.addstr(0, 0, reporting.status_report(jobs, n_cols, jobs_height, 
  File "/home/chia/plotman-main/reporting.py", line 107, in status_report
    plot_util.time_format(j.get_time_iowait())
  File "/home/chia/plotman-main/job.py", line 282, in get_time_iowait
    return int(self.proc.cpu_times().iowait)
AttributeError: 'pcputimes' object has no attribute 'iowait'```

Should check the other calls to psutil as well.

**Comment by ctrlaltdel** [comment]
Hi @ericaltendorf,

Thanks for this useful tool!

For the record, everything else works fine on Ubuntu 20.04 using package python3-psutis version 5.5.1-1ubuntu4 and the following hack.

```diff
     def get_time_iowait(self):
-        return int(self.proc.cpu_times().iowait)
+        return 0
```

Happy to make a proper PR for this if useful.

toastygunn:
My neighbor who I got into chia recently found the root cause of this issue. You need psutil ver 5.8 or better.
On his plotter which generated this error message, and my newly rebuild farmer which I decided could also do a little bit o' plotting before mainnet, you need psutil ver 5.6 or better (we installed 5.8).

$pip3 freeze
On problematic systems, this reports psutil==5.5.1

To install the newer ver, you need to type:
$pip3 install psutil==5.8

I dunno where the old ver came from, maybe apt get python-psutil? 
It doesn't matter b/c the problem is solved now for us by upgrading psutil without having to hack the job.py with a fake integer.

PR #53, opened 2021-03-24
#!/usr/bin/env python3

altendky:
This allows use in an activated venv, for example.


altendky:
The files where the `#!` is removed entirely didn't seem to 'do' anything beyond define functions and classes, so there's not much need for them to be directly runnable.  Unless I'm missing some bit in them.  I updated the execute permissions to match the presence or lack of a `#!`.

This follows helping someone in `#plotman` figure out getting latest `psutil` as noted in https://github.com/ericaltendorf/plotman/issues/38.

altendky:
Longer term, `entry_points` `console_scripts` is a nice part of an installable package with directly runnable files on all systems.

https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/

Though these days generally declaratively in a `setup.cfg` instead.

https://github.com/altendky/qtrio/blob/f77f7e009ce7f3985cbf7d4648ec0160096a55fc/setup.cfg#L52-L54
```ini
[options.entry_points]
console_scripts =
    qtrio = qtrio._cli:cli
```

ericaltendorf:
Looks good, thanks for the cleanup.  Will look at `entry_points` and `console_scripts`.

Changing to `development` base branch to merge.

PR #54, opened 2021-03-26
Pytest migrate

ericaltendorf:


altendky:
Maybe help GitHub reconsider this PR.

manager_test.py
```diff
@@ -1,82 +1,79 @@
...
-        self.sched_cfg = {
-                'tmpdir_stagger_phase_major': 3,
-                'tmpdir_stagger_phase_minor': 0,
-                'tmpdir_max_jobs': 3 }
-        self.dir_cfg = {
-                'tmp_overrides': {
-                        '/mnt/tmp/04': {
-                                'tmpdir_max_jobs': 4 }}}
+@pytest.fixture
+def sched_cfg():
```
altendky, line 11: This is perfectly correct.  But maybe you go and write a test like this.

```python
def test_stuff():
    assert len(sched_cfg) == 3
```

And oops, you are accessing the fixture function instead of having pytest pass you the fixture value because you forgot to put `sched_cfg` into your test function's parameter list.

You can avoid this by having your fixtures in another file.  Sometimes they are common and this makes sense.  I also end up doing this.

```python
@pytest.fixture(name='sched_cfg')
def sched_cfg_fixture():
```

Not particularly strongly suggesting a change here.  Just exposing a couple aspects of fixtures.

plot_util_test.py
```diff
@@ -1,77 +1,58 @@
 #!/usr/bin/env python3
```
altendky, line 2:
```suggestion
```

plot_util_test.py
```diff
@@ -1,77 +1,58 @@
...
-                plot_util.list_k32_plots('/t/') )
-
-
-if __name__ == '__main__':
-    unittest.main()
+def test_human_format():
+    assert (plot_util.human_format(3442000000, 0) == '3G')
+    assert (plot_util.human_format(3542000, 2) == '3.54M')
+    assert (plot_util.human_format(354, 0) == '354')
+    assert (plot_util.human_format(354, 2) == '354.00')
```
altendky, line 14: I dunno...  so the repetition isn't particularly an issue here but if the first fails you get no other feedback from which to derive a pattern of failure.

```suggestion
@pytest.mark.parametrize(
	argnames=['inhuman', 'human'],
	argvalues=[
	    [[3442000000, 0], '3G'],
	    [[3542000, 2], '3.54M'],
	    [[354, 0], '354'],
	    [[354, 2], '354.00'],
    ],
)
def test_human_format(inhuman, human):
	assert plot_util.human_format(*inhuman) == human
```

Also, is [humanize](https://python-humanize.readthedocs.io/en/latest/filesize/) what you want?  I didn't dig into MiB vs. whatever, maybe it isn't.  Also related, [pint](https://pint.readthedocs.io/en/stable/) for more general units handling.

plot_util_test.py
```diff
@@ -1,77 +1,58 @@
...
+            [ [ 0, 3, 6 ],
+              [ 1, 4, 7 ],
+              [ 2, 5, 8 ] ] )
+    assert (plot_util.column_wrap(list(range(3)), 1, filler='--') ==
+            [ [ 0 ],
+              [ 1 ],
+              [ 2 ] ] )
+
+# fs is a fake filesystem fixture magically supplied by pyfakefs
+def test_list_k32_plots(fs):
```
altendky, line 47: Not necessarily a thing to change now.  But, when you want type hinting it can replace the comment and also improve IDE completion and type checking.

```suggestion
def test_list_k32_plots(fs: pyfakefs.fake_filesystem.FakeFilesystem):
```

altendky (changes requested):

plot_util_test.py
ericaltendorf: that's exactly what i want.  i need to go learn about type hinting

manager_test.py
ericaltendorf: Thanks.  My brain is pretty full, but I'll try to register this for the future.

plot_util_test.py
ericaltendorf: regarding the libraries, they might be what i want.  let's figure out how to have a sane install process before i go adding more dependencies, though.

re: the other, yeah I didn't think about that.  I'm used to "expects" where you can list a bunch of them in a row.  I forgot switching to assert means it will bail on the first failure.  Is there a way to avoid that with pytest?  What you suggest could work, but i have a preference for very simple test code.

manager_test.py
ericaltendorf: What's the second code example demonstrating?  I don't think I understand.

altendky: The name of the fixture function in the second code block is `sched_cfg_fixture` so when you make the error in the first code block you'll get a straight `NameError` as you might 'normally expect' rather than `TypeError: object of type 'function' has no len()`.

plot_util_test.py
altendky: https://pypi.org/project/expects/?

plot_util_test.py
altendky: Probably don't close conversations I should look at.  Like here I'm scrolling through and this is all I see.

image

plot_util_test.py
altendky: Hmm...  I can certainly agree that my suggestion is not beautiful.  I don't offhand know an answer but did search for 'delayed assert' and lo and behold that was actually a relevant set of terms.  Found a few other related things.  Haven't dug into which would be 'best' or how great they really are.

https://pypi.org/project/delayed-assert/
https://pypi.org/project/pytest-assume/
https://github.com/pytest-dev/pytest/issues/1367

manager_test.py
ericaltendorf: Oh, the second is just a naming convention?

plot_util_test.py
ericaltendorf: Hmm, ok.  Something to look into.

manager_test.py
altendky: Along with the ability to set the fixture name via a parameter to the decorator rather than only by the name of the function itself.

ericaltendorf: OK.  I will have to read more about this one.

ericaltendorf:
Alright -- is this one good enough to go?

altendky (approved):
Sorry, I forgot there was an actual approval to be done...  Next step would be CI I guess?  Make sure that what you have here is being checked.  Not sure how you want to approach it but I just did up https://github.com/Parquery/pylddwrap/pull/16 this week.  It's somewhere in between a minimal but useful implementation and the full on setups for crossplatform auto-publishing etc.  I do tend to be more verbose than necessary to retain consistency across jobs in a project and across projects I set up.

PR #58, opened 2021-03-30
allow n plots to run before milestone, update config.yml

jianhaozhou:


config.yaml
```diff
@@ -69,12 +69,15 @@ directories:
...
-        # in sequence.
+        # Run a job on a particular temp dir only if the number of existing jobs
+        # before tmpdir_stagger_phase_major tmpdir_stagger_phase_minor
+        # is less than tmpdir_stagger_phase_limit.
+        # Phase major corresponds to the plot phase, phase minor corresponds to
+        # the table or table pair in sequence, phase limit corresponds to
+        # the number of plots allowed before [phase major, phase minor]
         tmpdir_stagger_phase_major: 2
         tmpdir_stagger_phase_minor: 1
+        tmpdir_stagger_phase_limit: 2
```
ericaltendorf, line 80: I guess a value of '2' in the example config is more illustrative of how one might use this, but I expect '1' to be the more common setting.  Should we make '1' the default in the example config?

jianhaozhou: Right, 1 makes more sense as the default. I'll go ahead and update it.

jianhaozhou:
@ericaltendorf I've updated the PR.

manager.py
```diff
@@ -52,7 +52,8 @@ def phases_permit_new_job(phases, d, sched_cfg, dir_cfg):
 
     milestone = ( sched_cfg['tmpdir_stagger_phase_major'],
                   sched_cfg['tmpdir_stagger_phase_minor'] )
-    if len([p for p in phases if p < milestone]) > 0:
+    max_allowed_before_milestone = sched_cfg['tmpdir_stagger_phase_limit']
```
ericaltendorf, line 55: One effect of this is that people already running plotman with a customized config.yaml will start to crash when they pick up this code, until they add the new line to their config.  Should we make this config value explicitly optional?

What we really need is some more managed install and upgrade process.  @altendky FYI

ericaltendorf (commented):
Thanks!  Was in the middle of leaving another comment too.

manager.py
jianhaozhou: Nice catch! Updated the PR.

ericaltendorf (approved):

ericaltendorf:
thanks for the contribution!

Issue #41, opened 2021-03-06
Parse command lines of existing jobs more robustly


ericaltendorf:
E.g.:
```
$ ./plotman.py interactive
Warning: unrecognized args: -k32 -n1
Warning: unrecognized args: -t/tmp -2/tmp2
Warning: unrecognized args: -d/dst -b6000
Warning: unrecognized args: -u128 -r3
...
```
Should probably use a standard library for parsing the args. :)

PR #59, opened 2021-03-31
quick fix for #41 which can handle the no-space case for known arg keys.

jkbecker:
As discussed on keybase.

This breaks up known "chia plot create" arguments when they don't contain a space character between key and value, to prevent faulty arg handling by the Job class.

Example:
```python
>>> cmdline = ['/chia-blockchain/venv/bin/python3.8', 'venv/bin/chia', 'plots', 'create', '-n1', '-k32', '-r12', '-t/plots/tmp', 
'-2/plots/tmp', '-d/plots', '-b9216']
```
Would usually result in faulty parsing and a `StopIteration` exception in job.py line 105:
```
Warning: unrecognized args: -k32 -r
Warning: unrecognized args: 12 -t
Warning: unrecognized args: /plots/tmp -2
Warning: unrecognized args: /plots/tmp -d
Warning: unrecognized args: /plots -b
```
i.e. completely mismatched argument keys/values.
Solution:
```python
>>> cmdline_argfix(cmdline)
['/chia-blockchain/venv/bin/python3.8', 'venv/bin/chia', 'plots', 'create', '-n', '1', '-k', '32', '-r', '12', '-t', '/plots/tmp', '-2', '/plots/tmp', '-d', '/plots', '-b', '9216']
```

### DIFF
--- a/job.py
+++ b/job.py
@@ -33,6 +33,17 @@ def is_plotting_cmdline(cmdline):
         and 'create' == cmdline[3]
     )
 
+# This is a (temporary?) fix for https://github.com/ericaltendorf/plotman/issues/41
+def cmdline_argfix(cmdline):
+    known_keys = 'krbut2dne'
+    for i in cmdline:
+        if i[0]=='-' and i[1] in known_keys and len(i)>2:
+            key = i[0:2]
+            val = i[2:]
+            cmdline[cmdline.index(i)] = key
+            cmdline.insert(cmdline.index(key)+1, val)
+    return cmdline
+
 # TODO: be more principled and explicit about what we cache vs. what we look up
 # dynamically from the logfile
 class Job:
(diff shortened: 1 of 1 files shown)


job.py
```diff
@@ -33,6 +33,17 @@ def is_plotting_cmdline(cmdline):
...
 
+# This is a (temporary?) fix for https://github.com/ericaltendorf/plotman/issues/41
+def cmdline_argfix(cmdline):
+    known_keys = 'krbut2dne'
+    for i in cmdline:
+        if i[0]=='-' and i[1] in known_keys and len(i)>2:
+            key = i[0:2]
+            val = i[2:]
+            cmdline[cmdline.index(i)] = key
+            cmdline.insert(cmdline.index(key)+1, val)
```
altendky, line 44: I would tend to avoid the mutation of the `cmdline` argument here.  This could instead either create a new list to return or yield out the elements as a generator.  Mutating arguments can be confusing and surprising to callers.  Also, this loop is mutating the list it is iterating over.  I think that the next iteration will actually be checking the newly-inserted element.  This is probably not a problem in most cases, but doesn't seem likely to be good or wanted.

If the mutation were to be kept, the loop could be `for index, element in enumerate(cmdline):` to avoid the multiple `cmdline.index()` calls that are searching through the list (and hopefully finding the correct element and not some other that matches).

jkbecker: True, it's a bit messy. Will implement it differently when I redo this on the development branch.

jkbecker:
Closing this, will resubmit something based off of the development branch that considers your input above.

PR #60, opened 2021-03-31
Handle spaceless args like '-k32', not just '-k 32'

jkbecker:
Redid #59 based on the development branch, and improved implementation based on @altendky's feedback.
Fixes #41.

---

This breaks up known `chia plot create` arguments when they don't contain a space character between key and value, to prevent faulty arg handling by the Job class.

Example:
```python
>>> cmdline = ['/chia-blockchain/venv/bin/python3.8', 'venv/bin/chia', 'plots', 'create', '-n1', '-k32', '-r12', '-t/plots/tmp', 
'-2/plots/tmp', '-d/plots', '-b9216']
```
This would usually result in faulty parsing and a `StopIteration` exception in job.py line 105:
```
Warning: unrecognized args: -k32 -r
Warning: unrecognized args: 12 -t
Warning: unrecognized args: /plots/tmp -2
Warning: unrecognized args: /plots/tmp -d
Warning: unrecognized args: /plots -b
```
i.e. completely mismatched argument keys/values.
Solution:
```python
>>> cmdline_argfix(cmdline)
['/chia-blockchain/venv/bin/python3.8', 'venv/bin/chia', 'plots', 'create', '-n', '1', '-k', '32', '-r', '12', '-t', '/plots/tmp', '-2', '/plots/tmp', '-d', '/plots', '-b', '9216']
```

PR #61, opened 2021-03-31
Setup

jkbecker:
A setuptools-based installation script as well as the required re-structuring of the entire project.

See updated README.md for the new installation instructions once this is in effect.


jkbecker:
Just some mental notes of what doesn't work yet

- [ ] Test setup
- [ ] getting the config file from the correct location based on where the package installed it
- [ ] Clarify if MANIFEST.in is even needed - the files it contains don't seem to be copied into the installed package
- [ ] Probably more

setup.cfg
```diff
@@ -0,0 +1,65 @@
...
+    Development Status :: 3 - Alpha
+    Environment :: Console :: Curses
+    Intended Audience :: Developers
+    Intended Audience :: System Administrators
+    Intended Audience :: Information Technology
+    License :: OSI Approved :: Apache Software License
+    Natural Language :: English
+    Operating System :: OS Independent
+    Programming Language :: Python :: 3
+    Topic :: Office/Business :: Scheduling
```
altendky, line 24: I think this is maybe for other forms of scheduling.  `:]`

altendky (commented):
Overall, it seems a pretty tidy transition as it should be.  Thanks!  Don't let the quantity of minor comments etc express otherwise.  And yeah, in work etc.  I prefer to share early and often in case there are significant issues that should be addressed early.  Handily, I don't think there are any here.

I'll note here that this includes the changes in #60.  If this is merged first it is accepting both changes.  I don't have a problem with #60, just pointing it out.

`MANIFEST.in` is for the sdist, IIRC.  If you specify `include_package_data = True` then it will make it into the wheel as well and thus the installation.  IIRC...  I think this may also replace `[options.data_files]`.  But it's bin awhile since I've fiddled with this and it's one of the areas where there are many ways to do it and most have gotchas.

Maybe https://pypi.org/project/appdirs/ and a CLI override option for config file locations?  Though this strikes me as maybe a separate issue unless installation breaks the existing functionality?  I should go see what it is.

Maybe `src/plotman/plotman.py` should be `core.py` or something?  Module paths like `plotman.plotman` strike me as a bit odd.

I would avoid encouraging people to activate the environment when running.  There's plenty of room for personal preference on whether to activate or not but from the perspective of supporting people it ends up adding hidden global state in `$PATH` that you have to debug.  Whereas if people send you a traceback with `venv/bin/plotman` then you know they are running the plotman in the env.

Looking good for a first share.  I think this is one of the first things I poked @ericaltendorf about getting done with plotman.  Good to see it happening!

setup.cfg
```diff
@@ -0,0 +1,65 @@
...
+version = file: VERSION
+author = Eric Altendorf
+author-email = PLACEHOLDER
+home-page = https://github.com/ericaltendorf/plotman
+description = Chia plotting manager
+long-description = file: README.md
+long_description_content_type = text/markdown
+license = Apache 2.0
+license-file = LICENSE
+platform = any
```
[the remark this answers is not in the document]
jkbecker: Honestly I was working off a template and didn't really think about it. You're right, I have no idea how psutil and the plot spawning shenanigans work outside of Linux, should probably be explicitly limited to linux just to save some headache from people running it on win10 etc.

src/plotman/plotman.py
```diff
@@ -202,3 +198,5 @@ def get_term_width():
                     print('Resuming ' + job.plot_id)
                     job.resume()
 
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
jkbecker: sounds good

setup.cfg
jkbecker: I definitely meant it in this sense:
image

altendky:
@jkbecker, for getting the config from the 'right place', I think that can be a separate PR?  This is already a restructure so keeping it focused is more valuable than normal.  Also keeping it from getting held up on other changes is good.  Unless I'm misunderstanding and the config location is particularly affected by this?  From a quick search it seems to just be looking in cwd right now.

jkbecker:
The main reason I wanted to couple those topics is: Once plotman is properly installed, it can be launched with the `plotman` command from anywhere, but the `config.yaml` references in the code are dependent on the current working directory. Essentially this will break if you're in the wrong directory, and as soon as you have a globally available command, that's not intuitive.

That said, considering I'll probably work on that next anyways, we could separate it as well?

There are some considerations when it comes to "correct" config file location handling.

* Typically you'd want the default location to be in [appdir's](https://pypi.org/project/appdirs/)  `user_config_dir`
* To satisfy existing users we may want to attempt to read the `config.yaml` in the current directory with higher priority (?) (if it is in fact a plotman config.yaml - could be dangerous to just ingest any file with that name that happens to be in the working directory...
* With highest priority, `plotman` should also respect path overloading via a variable such as `PLOTMAN_CONFIG=~/config.yaml plotman interactive` if desired

Does this make sense? Am I missing a special case? Happy to deal with this in a separate PR if it makes the individual contributions more manageable, either way is fine for me.

altendky:
Agreed on all points.  I'm not sure if you said this indirectly or not, but `config.yaml` may just be a bad name at this point.  Certainly understandable how it came to be though.  `:]`  I'd tend to separate the two just to avoid holding one up on the other and to have simpler to track reviews.  Won't be making a fuss either way though.

README.md
```diff
@@ -166,15 +166,25 @@ mode or the command line mode.
...
 
-python3 -m pip install psutil pyfakefs pyyaml texttable
-```
+       git clone https://github.com/ericaltendorf/plotman.git -b development
+3. Enter the Plotman directory
+
+       cd plotman
+4. Install plotman using pip:
+
+       pip install .
```
altendky, line 188: I don't know what workflow we want to encourage in the interrim here until plotman is on PyPI.  We could skip the git cloning and just `pip install git+https://github.com/ericaltendorf/plotman@development` or such.

src/plotman/job.py
```diff
@@ -1,18 +1,19 @@
...
 import threading
 import time
-import psutil      # apt-get install python-psutil
-import random
-import sys
+from datetime import datetime
+from enum import Enum, auto
+from subprocess import call
+
+import psutil  # apt-get install python-psutil
```
altendky, line 15:
```suggestion
import psutil
```

altendky (changes requested):
I'll commit the minor tweaks tomorrow after reviewing my own review at a more appropriate hour.  @ericaltendorf do you want to discuss the effects of this on users before merging to develop?  No more launching as `./plotman.py` and such.  Or, is the addition to the README sufficient?

Thanks again @jkbecker, definitely happy to have this change.  `:]`

setup.cfg
```diff
@@ -0,0 +1,62 @@
+[metadata]
+name = plotman
+version = file: VERSION
+author = Eric Altendorf
+author-email = PLACEHOLDER
```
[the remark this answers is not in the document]
ericaltendorf: Yeah, let's omit the email for now.  It would likely only give people a false hope that I'd have time to fix things for them. :)

ericaltendorf:
I think the addition to the README should be sufficient, along with an announce on the plotman keybase channel.  I guess improving the `config.yaml` situation will be a separate bit of work?

(FWIW what I've been doing so far is creating a sibling directory (e.g., `plotman-run/`) to the `plotman/` git local repo, putting a symlink to `../plotman/plotman.py` in there and a copy of `config.yaml` that I modify with my local settings.)

README.md
jkbecker: Fine with me, definitely easier with fewer steps and no explicit git knowledge required! Haven't tested it but it might need a `#egg=plotman` at the end of the URL? At least I've seen that construct in other similar cases before.

jkbecker: Ok, I reworked this in commit 8b06c62, please check if this is resolved.

README.md
```diff
@@ -166,15 +166,23 @@ mode or the command line mode.
...
 There are many bugs and TODOs.
 
-## Installation
+Plotman will always look in the current directory for the `config.yaml` file
+([see also](https://github.com/ericaltendorf/plotman/pull/61#issuecomment-812967363)).
 
-This program requires `psutil`, `pyfakefs`, `pyyaml` and `texttable`.
+## Installation
 
 Installation for Ubuntu 20.04:
```
altendky, line 174: The only bit even Linux specific at this point is the use of `.` for the activation.  Though it is probably just as well to say Linux since we don't support Windows presently.

```suggestion
Installation for Linux:
```

README.md
altendky: Seems good, thanks.

altendky (approved):

altendky:
I guess I may as well draft the keybase message here.  Suggestions and corrections are of course appreciated.  Or, you could handle this if you want, @jkbecker.  `:]`

---

Hi all.  @jkbecker has kindly reworked the project structure of plotman for us so that it will be a properly installable Python package.  This is being merged into the `development` branch.  The new setup process has been documented in the readme.  While this specific change does not affect how plotman finds the `config.yaml` it would be understandable for some to be unsure where to put it.  plotman still looks for `config.yaml` in the current working directory.  This is the directory you are in in your shell when you run the program.  You can keep your config wherever you wish.

Now that I've mentioned how the `config.yaml` is found presently, I'll go ahead and note that there has been some talk of changing that and the topic is ticketed in #84.

https://github.com/ericaltendorf/plotman#installation
https://github.com/ericaltendorf/plotman/issues/84

jkbecker:
Looking good @altendky, thank you! I prefer not to announce it myself, as I'm not previously associated with the project. Don't want this to look like an intrusion ;)

❗  One detail just came to mind: If people just read the README on GitHub and `pip install` as suggested, they will actually have **no** config.yaml on their system at all! We're not covering this edge case.

Or are we assuming (for now), that people installing the development branch are existing users and somewhat more experienced with how Plotman works? At the very least, #84 should be addressed before merging to main.

altendky:
It's not like I have some Long standing involvement with plotman either.  `:]`  But anyways, good points.  I'll add some wording both places on how to get a copy of the configuration for now.  I'm good with also getting some configuration location changes in before a 'release' (I think that's what merging to `main` is?).

jkbecker:
Hm, random idea: Maybe it would be a good idea to merge the current development branch progress to main, **before** merging setup into development?

I feel like there are a lot of small improvements in development that could be helpful **now**, people are complaining about bad GUI plot handling almost every other hour for example. Merging setup onto development would just make merging develop on main a much bigger change, right?

altendky (changes requested):
The request for more commentary on `config.yaml` is worthy of being a requested change I think.  Even though I plan on doing it.

https://github.com/ericaltendorf/plotman/pull/61#issuecomment-813808315

altendky (approved):

PR #66, opened 2021-04-01
Add GitHub Actions and tox

altendky:


altendky:
This is pretty minimal, but does run the tests.  Maybe double check it runs all of the ones you want...  The commented out stuff should come back with #61 so I'm being a bit dirty and leaving it there.  We can add various checks separately.

altendky:
I may as well link https://github.com/Parquery/pylddwrap/pull/16#issuecomment-808311535 where there is a little commentary about the structure.  Or we can just chat about it as it interests you.  This is pretty much copied from pylddwrap since that is my most recent effort and also a Linux-only CI setup like I think plotman can start with.  Correct me if you expect it to work on other platforms and we can add those in pretty easily.

jkbecker:
Looks nice! I feel like this should be merged before #61, then I can pull it and make sure it also works with the directory structure I built there.

Re: tox: The official blockchain [requires Python 3.7](https://github.com/Chia-Network/chia-blockchain/wiki/INSTALL), is there a reason to bother with versions below that for plotman?

altendky:
I tend towards supporting all supported Pythons (and in legacy cases also Python 2.7 when it isn't too onerous).  That's certainly a reasonable argument about 3.7 though.  On the flip side, didn't you just run plotman on a system without chia-blockchain?  `:]`  I'm trying to think what the big reasons are for requiring 3.7 in general.  I think asyncio grew some pleasant stuff.  `from __future__ import annotations` would be relevant for type hinting (you don't have to quote forward references).  Probably something I'm forgetting...  Anyways, it's probably not a big deal either way.  @ericaltendorf can flip the coin.  (or pretend they did and just tell us the answer)

ericaltendorf (approved):

altendky:
Oh yeah, [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) got some nice tweaks too.

> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines. Added the capture_output parameter.

You may yet get me on the 3.7+ bandwagon.  `:]`

jkbecker:
Wait until you find out about the Walrus operator (>=3.8) and want to use it everywhere (:

PR #74, opened 2021-04-03
feat: handle job --long-arguments

altendky:
Draft for:
- [x] Self review
- [ ] ~~Maybe some tests...~~
  - Maybe we should get coverage checks setup before mandating tests be added

This is really a stop-gap, similar to #60, for generally more complete CLI parsing.  Also note #41.


altendky:
Side note, perhaps launching jobs should use `--long-options` for enhanced legibility.  At least for me, I tend to do long options any time I'm not typing at the shell repeatedly.

jkbecker (approved):
Yeah looks good to me. All signs point toward click medium-term, though.

PR #76, opened 2021-04-03
feat: treat ctrl+c as a quit request

altendky:
Draft for:
- [x] ~~I really don't know if this is a good place to put this or if it should be much broader.~~
  - It should be much broader but this is a simple step in the right direction.

```python-traceback
  File "/farm/venv/bin/plotman", line 33, in <module>
  File "/farm/plotman/src/plotman/plotman.py", line 125, in main
    interactive.run_interactive()
  File "/farm/plotman/src/plotman/interactive.py", line 328, in run_interactive
    curses.wrapper(curses_main)
  File "/home/altendky/.pyenv/versions/3.9.2/lib/python3.9/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/farm/plotman/src/plotman/interactive.py", line 301, in curses_main
    key = stdscr.getch()
KeyboardInterrupt
```


jkbecker:
I think this is a good & simple fix.
Slightly off-topic, but in the realm of clean quitting, I had my caps locked yesterday and capitalized `Q` does not trigger quit. that seems like an easy thing to add as well?

ericaltendorf:
sure, sgtm

altendky:
So, we want this as-is as an improvement without addressing less-likely cases?  I think I saw a ctrl+c hit not-here, though there's good odds it will.  I hadn't looked at the rest of the code to get better coverage.  Maybe a real fix would look like a `SIGINT` handler setting a global variable that we check 'occasionally' and trigger a quit.  Need to make sure a second one still allows for a normal uncontrolled failout.

jkbecker:
Honestly I've never worked with curses so I'm not sure if a SIGINT could be captured and not properly handled in there somewhere? No experience. We could add the current code because it's fairly simple and should have no side-effects otherwise, and if it turns out that in some places ctr-c still doesn't get captured, we can expand using a more complex/comprehensive construct?

altendky:
@jkbecker, I forget the details of how GitHub handles this...  Are you able to submit an approving review?

jkbecker:
@altendky I'm not being shown any way to act as a reviewer here. Maybe I need to be added to the project and/or tagged in the PR as a reviewer somehow?

I'm seeing this on the side and all I can do is comment:
image

altendky:
I think you can do it over on the `Files changed` 'tab'.

image

jkbecker (approved):
Hah, indeed. Weird UX

PR #85, opened 2021-04-05
fix: acknowledge any chia command instead of requiring venv/bin/chia

altendky:
Maybe someday we use pathlib everywhere, but for now this cuts it.

### DIFF
--- a/job.py
+++ b/job.py
@@ -26,7 +26,7 @@ def is_plotting_cmdline(cmdline):
     return (
         len(cmdline) >= 4
         and 'python' in cmdline[0]
-        and 'venv/bin/chia' in cmdline[1]
+        and cmdline[1].endswith('/chia')
         and 'plots' == cmdline[2]
         and 'create' == cmdline[3]
     )


job.py
```diff
@@ -26,7 +26,7 @@ def is_plotting_cmdline(cmdline):
     return (
         len(cmdline) >= 4
         and 'python' in cmdline[0]
-        and 'venv/bin/chia' in cmdline[1]
+        and cmdline[1].endswith('/chia')
```
ericaltendorf, line 29: if we're generalizing, does `chia` need to be path-specified?  should it be `== 'chia' or endswith('/chia')` ?

altendky: Now that I've written all this, hopefully I understood the question correctly...

Note the exploratory console output at the end.

I think you will end up with a full path for 'normal' invocations of Chia via its entry point.  The first entry in the command line in the psutil output I think is based on the `#!` of the entry point at `/farm/venv/bin/chia`, in my case.  But, it was discovered at `/home/chia/.local/bin/chia` as a symlink `/farm/venv/bin/chia`.  Seemingly, Linux decides to use the unresolved symlink path as the argument to the `#!` provided Python command.  Given that we end up with a full path in the reported command line despite never having specified it, it seems like it will 'always' be a full path which means there will be a `/` before the file name.  But, I don't really know the rules around this.  Just reaching the conclusion by these observations.

If we switch over to `pathlib` this would be written as `pathlib.Path(cmdline[1]).name == 'chia'` and that would work similarly to your proposed check.  So, I think `== 'chia'` is not wrong to match but it seems like it will be superfluous in the case of normal Chia usage.

But...  there's another case here sadly.  It seems that they've set it up so you could in fact run `python path/to/chia.py` or `python chia.py` if you are down inside their package.  `:[`  In which case none of this will catch it since it would be something like `/whatever/venv/lib/python3.9/site-packages/src/cmds/chia.py` (`src` will become `chia` or something once they fix their top level import package name) or just `chia.py`.

I think it may be reasonable to scope this, at least for now, to 'processes started by plotman or otherwise started via the entry point'.  If you do want to try to catch the other mentioned cases then...  Hmm...  This alone could catch people having written their own `chia.py` but unless they laid out the same subcommands etc the other bits of the check would result in skipping over that process.  So, probably mostly safe from false positives.

```suggestion
        and pathlib.Path(cmdline[1]).name in {'chia', 'chia.py'}
```

At some point of detail here, I don't want to be writing a specific `/` what with Windows etc.  Though I'm not sure how the whole `#!` bit works out in Windows.  I mean, I know a bit about how it ends up being used (what with Windows not supporting `#!` like Linux) but I don't know any details about the resulting process' command line.

```console
$ venv/bin/python -c 'import psutil; print(psutil.Process(2138191).cmdline())'
['/farm/venv/bin/python3', '/home/chia/.local/bin/chia', 'plots', 'create', '-k', '32', '-r', '8', '-u', '128', '-b', '4000', '-t', '/farm/yards/900', '-d', '/farm/wagons/800']
```

```console
$ ls -l /home/chia/.local/bin/chia
lrwxrwxrwx 1 chia chia 19 Apr  4 20:59 /home/chia/.local/bin/chia -> /farm/venv/bin/chia*
```

```console
$ ls -l /farm/venv/bin/chia
-rwxrwxr-x 1 chia chia 218 Apr  4 20:57 /farm/venv/bin/chia*
```

```console
$ head /farm/venv/bin/chia
#!/farm/venv/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from src.cmds.chia import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

https://github.com/ericaltendorf/plotman/blob/b8918b4fcb96fac501352c76f864559f139aaccc/manager.py#L115

https://github.com/Chia-Network/chia-blockchain/blob/ad7d7e0cced7f2f6deddc9e006dbaeee6dab8f66/src/cmds/chia.py#L80-L81
```python
if __name__ == "__main__":
    main()
```

ericaltendorf: OK.  I have no idea how those values get set, or how consistent they are across platforms.  I'm OK with any approximate heuristic because I doubt there's a really solid way to do this.  The "right" solution to most of these questions would actually be some kind of real integration with the core chia code base.

Also, the length of your reply leaves me a bit concerned about how much time you're spending on this...?  :)

ericaltendorf (approved):

Issue #77, opened 2021-04-03
config should be validated (like a schema check)


altendky:
User reported some confusing issues with paths.  Turns out they had `tmp: /my/path` instead of `tmp: [/my/path]` (or a multiline yaml list).  The config file should be checked against a schema before we go using it.  The following code expects a list but is just fine taking a string `'/my/path'` and turning it into a bunch of temporary directories like `['/', 'm', 'y', '/', 'p', 'a', 't', 'h']`.  This bug (we guess) also resulted in temporary files getting dumped in `/` which strikes me as extra bad.

```yaml
        tmp: /root/venvChia/plots/temp1
```

https://github.com/ericaltendorf/plotman/blob/31ccb8f9c48b30b80a40c6bc3af167edf6b3351e/manager.py#L79-L80

<details> <summary>complete config file</summary>

```yaml
# Where to plot and log.
directories:
        # One directory in which to store all plot job logs (the STDOUT/
        # STDERR of all plot jobs).  In order to monitor progress, plotman
        # reads these logs on a regular basis, so using a fast drive is
        # recommended.
        log: /root/venvChia/plotman/logs
 
        # One or more directories to use as tmp dirs for plotting.  The
        # scheduler will use all of them and distribute jobs among them.
        # It assumes that IO is independent for each one (i.e., that each
        # one is on a different physical device).
        #
        # If multiple directories share a common prefix, reports will
        # abbreviate and show just the uniquely identifying suffix.
        tmp: /root/venvChia/plots/temp1
 
 
 
        # Optional: tmp2 directory.  If specified, will be passed to
        # chia plots create as -2.  Only one tmp2 directory is supported.
        # tmp2:
 
        # One or more directories; the scheduler will use all of them.
        # These again are presumed to be on independent physical devices,
        # so writes (plot jobs) and reads (archivals) can be scheduled
        # to minimize IO contention.
        dst: /root/venvChia/plots/plot4
 
        # Archival configuration.  Optional; if you do not wish to run the
        # archiving operation, comment this section out.
        #
        # Currently archival depends on an rsync daemon running on the remote
        # host, and that the module is configured to match the local path.
        # See code for details.
        # archive:
        #        rsyncd_module: plots
        #        rsyncd_path: /plots
        #        rsyncd_bwlimit: 80000  # Bandwidth limit in KB/s
        #        rsyncd_host: myfarmer
        #        rsyncd_user: chia
 
 
# Plotting scheduling parameters
scheduling:
[...]

PR #88, opened 2021-04-07
Local archive mode (fixes #6)

### LINKED ISSUE #6: Enable local archiving (and improve archiving design)
Currently, the archiving process assumes archiving to a remote host using `rsync`.

We should allow multiple transports.  At least, rysnc between local dirs.

The main complication here is that we not only need to transfer files, but also to check free disk space on the target directories.  Currently this is implemented by remotely executing the `df` command on the remote host.  So we need a clean and robust way to check free space that is coordinated with the file transport mechanism.

There's an additional complication because when transferring via rsync to a remote rsync daemon, we see virtual paths as exported by the rsync daemon configured modules, but when we execute `df`, we view paths as they exist natively on the remote host.  We currently have a fairly crufty mechanism for mapping between the two views of paths.  This mechanism needs to be robust, and to generalize across whatever transport mechanisms we implement.

jkbecker:
Hey,

I needed a local archive mode, so I implemented one.

What can I say,
image


config.yaml
```diff
@@ -76,6 +87,7 @@ directories:
                 #   index: 0
 
 
+
```
altendky, line 90: Looks like double blanks elsewhere but triple here.

```suggestion
```

config.yaml
```diff
@@ -60,11 +60,22 @@ directories:
         # host, and that the module is configured to match the local path.
         # See code for details.
         archive:
-                rsyncd_module: plots
+                mode: remote  # remote or local
+                rsyncd_module: plots  # ignored in local mode
+
+                # Note for local mode:
```
altendky, line 66: Maybe there should be a `local` mapping containing the local config rather than sharing `rsyncd_path`?  I guess local is still using rsync, not rsyncd.  Hmm...  I'm really not sure.  But, I'll leave this here to see how you feel.

In general, look at integrating the existing remote and new local comments with each other.

src/plotman/archive.py
```diff
@@ -49,8 +49,14 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = { }
-    df_cmd = ('ssh %s@%s df -aBK | grep " %s/"' %
-        (arch_cfg['rsyncd_user'], arch_cfg['rsyncd_host'], arch_cfg['rsyncd_path']) )
+    arch_mode = arch_cfg.get('mode', 'remote')
```
altendky, line 52: If this is consistent with other defaults, ok.  Though it really should be handled centrally so we don't accidentally default differently in different places.  But, any real fix of this system as a whole should be handled separately and is probably related to #77.

src/plotman/archive.py
```diff
@@ -49,8 +49,14 @@ def compute_priority(phase, gb_free, n_plots):
...
+    if arch_mode == 'remote':
+        df_cmd = ('ssh %s@%s df -BK | grep " %s/"' %
+            (arch_cfg['rsyncd_user'], arch_cfg['rsyncd_host'], arch_cfg['rsyncd_path']) )
+    elif arch_mode == 'local':
+        df_cmd = ('df -BK | grep " %s/"' % arch_cfg['rsyncd_path'] )
+    else:
+        raise KeyError(f'Archive mode must be "remote" or "local" ("{arch_mode}" given). Please inspect config.yaml.')
```
altendky, line 59: This will avoid accidentally invalid quoting by using `repr()`.  Though it also will give you single quotes usually which isn't what you had.  `:|`

https://docs.python.org/3.9/tutorial/inputoutput.html#formatted-string-literals

```suggestion
        raise KeyError(f'Archive mode must be "remote" or "local" ({arch_mode!r} given). Please inspect config.yaml.')
```

---

I personally don't use builtin exceptions and instead make my own.  I often make a base exception per project and inherit all others from that.  But, that's not a mandate here.  Plenty of opinions on that one.

src/plotman/archive.py
```diff
@@ -60,11 +66,17 @@ def get_archdir_freebytes(arch_cfg):
     return archdir_freebytes
 
 def rsync_dest(arch_cfg, arch_dir):
-    rsync_path = arch_dir.replace(arch_cfg['rsyncd_path'], arch_cfg['rsyncd_module'])
-    if rsync_path.startswith('/'):
-        rsync_path = rsync_path[1:]  # Avoid dup slashes.  TODO use path join?
-    rsync_url = 'rsync://%s@%s:12000/%s' % (
-            arch_cfg['rsyncd_user'], arch_cfg['rsyncd_host'], rsync_path)
+    arch_mode = arch_cfg.get('mode', 'remote')
```
altendky, line 69: Heh, alredy defaulting it twice.  Yep, at least make this get set in one place with `.setdefault()` or such after loading.  Though...  we load in two different places without a function so maybe we'll have to go fix that separately first.

src/plotman/archive.py
```diff
@@ -49,8 +49,14 @@ def compute_priority(phase, gb_free, n_plots):
...
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = { }
-    df_cmd = ('ssh %s@%s df -aBK | grep " %s/"' %
-        (arch_cfg['rsyncd_user'], arch_cfg['rsyncd_host'], arch_cfg['rsyncd_path']) )
+    arch_mode = arch_cfg.get('mode', 'remote')
+    if arch_mode == 'remote':
+        df_cmd = ('ssh %s@%s df -BK | grep " %s/"' %
+            (arch_cfg['rsyncd_user'], arch_cfg['rsyncd_host'], arch_cfg['rsyncd_path']) )
+    elif arch_mode == 'local':
+        df_cmd = ('df -BK | grep " %s/"' % arch_cfg['rsyncd_path'] )
```
altendky, line 57: There's some duplication here that makes me itchy.  But I think the piping and grepping add some complexity here.  In one case the `df` and `grep` are on different systems and on the other they are on the same system.  But I think really I'd like to see the pipe dropped entirely.  `[line for line in lines if f" {arch_cfg['rsyncd_path']}/" in line]` is a small price to pay for not needing to use a shell when subprocessing.  (or something fancier maybe if I thought about what was actually being done)  Not to mention that we had some hazards there since we weren't escaping the path for `grep`...

altendky (changes requested):
Ok, so two concerns here lead off to other tickets that would be nice to take care of before this.  Sometimes you feel like doing that, sometimes you don't...  if you don't want to do it I think I'll reserve a day or two for myself to hopefully get to them.  If neither of us do then yeah, we can swing back around to dealing with them later.

1. There should be a single function holding the code for loading the configuration file.  Presently there are two places it is loaded.  This function is where defaults should be set.
2. The `grep`ing of the `df -aBK` output should just be done in Python and the shell removed from the subprocessing.

altendky:
I'm going to withdraw my request to hold this up on the other changes.  I don't think merging this first makes it any harder to get everything done so no reason to wait.  Sorry for that.  Anyways, let me take another pass here and resolve those discussions.

src/plotman/archive.py
altendky: No matter which comes first but there's https://github.com/ericaltendorf/plotman/pull/98 and wherever that comes together with this we can make get this default set at a single point.

src/plotman/archive.py
altendky: I can deal with the `grep` removal separately.  No need to hold this up on that.

altendky:
@jkbecker, where are we at here?  So there's some catch up to do with other changes, like the configuration stuff.  I think I wanted the comments in the configuration file to be merged with the existing ones.  Maybe a separate local archive config section?  Or, maybe this is now caught up in the general 'allow arbitrary archiving actions'?  https://github.com/ericaltendorf/plotman/issues/110

jkbecker:
Yeah, I feel like it's caught up in the broader discussion, but it could be used as a transitional thing as long as it is disabled by default and opt-in?

I won't have the bandwidth to expand on this or make it a more  generally-applicable setup in the next few weeks to come.

src/plotman/configuration.py
```diff
@@ -46,6 +46,7 @@ class Archive:
     rsyncd_host: str
     rsyncd_user: str
     index: int = 0  # If not explicit, "index" will default to 0
+    mode: Optional[str] = 'remote'
```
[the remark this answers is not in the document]
altendky: `Optional` means 'can be `None`' not 'can be left out of config'.  Having a default value means 'can be left out of config'.

```suggestion
    mode: str = 'remote'
```

src/plotman/configuration.py
```diff
@@ -46,6 +46,7 @@ class Archive:
     rsyncd_host: str
     rsyncd_user: str
     index: int = 0  # If not explicit, "index" will default to 0
+    mode: str = 'remote'
```
altendky, line 49: So I think that maybe we can validate this here.

(imports would need to be added as well)
```suggestion
    mode: str = desert.field(
        default='remote',
        marshmallow_field=marshmallow.fields.String(
            validate=marshmallow.validate.OneOf(choices=['remote', 'local']),
        ),
    )
```

Probably don't bother for now, but having this be an enumeration might be good.  Making desert support `mode: typing.Literal['remote', 'local'] = 'remote'` would also be good.  But, oh well.  Another day.

Anyways, this can get us errors like the following at the time of load.

https://replit.com/@altendky/MicroSkeletalLinks-1
```python-traceback
  File "main.py", line 19, in <module>
    result = schema.loads('{"s": "blue"}')
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/marshmallow/schema.py", line 759, in loads
    return self.load(data, many=many, partial=partial, unknown=unknown)
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/marshmallow/schema.py", line 725, in load
    return self._do_load(
  File "/opt/virtualenvs/python3/lib/python3.8/site-packages/marshmallow/schema.py", line 907, in _do_load
    raise exc
marshmallow.exceptions.ValidationError: {'s': ['Must be one of: remote, local.']}
```

src/plotman/archive.py
```diff
@@ -77,8 +77,13 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = {}
-    df_cmd = ('ssh %s@%s df -aBK | grep " %s/"' %
-        (arch_cfg.rsyncd_user, arch_cfg.rsyncd_host, arch_cfg.rsyncd_path) )
+    if arch_cfg.mode == 'remote':
+        df_cmd = ('ssh %s@%s df -aBK | grep " %s/"' %
+            (arch_cfg.rsyncd_user, arch_cfg.rsyncd_host, arch_cfg.rsyncd_path) )
+    elif arch_cfg.mode == 'local':
+        df_cmd = ('df -BK | grep " %s/"' % arch_cfg.rsyncd_path )
```
altendky, line 84: I don't know if we want `-a` or not, but do we want local to be different than remote?

```suggestion
	just_df_cmd = 'df -aBK'
    if arch_cfg.mode == 'remote':
        df_cmd = ('ssh %s@%s %s | grep " %s/"' %
            (arch_cfg.rsyncd_user, arch_cfg.rsyncd_host, just_df_cmd, arch_cfg.rsyncd_path) )
    elif arch_cfg.mode == 'local':
        df_cmd = '%s | grep " %s/"' % (just_df_cmd, arch_cfg.rsyncd_path)
```

src/plotman/archive.py
```diff
@@ -77,8 +77,14 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = {}
-    df_cmd = ('ssh %s@%s df -aBK | grep " %s/"' %
-        (arch_cfg.rsyncd_user, arch_cfg.rsyncd_host, arch_cfg.rsyncd_path) )
+	just_df_cmd = 'df -aBK'
```
lopesmcc, line 80:
```suggestion
    just_df_cmd = 'df -aBK'
```

lopesmcc (changes requested):

src/plotman/archive.py
altendky: Quite so, thanks.

fihzy:
I stitched this local archiving pull request in to my setup and it didn't work. The box I was running it on was receiving rsync plots already from another node -  get_running_archive_jobs was seeing them and refusing to start a local archive. I don't know if this is already fixed somewhere in here, but if not - I tracked it down to dest = rsync_dest(arch_cfg, '/') where the / was matching the args for any running rsync process since they all start with a / . To fix it  changed that block like this (key part is replacing / with actual rsync path)

`if arch_cfg.mode == 'remote':
        dest = rsync_dest(arch_cfg, '/')
    elif arch_cfg.mode == 'local':
        dest = rsync_dest(arch_cfg, arch_cfg.rsyncd_path)
    else:
        raise KeyError(f'Archive mode must be "remote" or "local" ({arch_cfg.mode!r} given). Please inspect plotman.yaml.')
`

config.yaml
jkbecker: I have implemented something like this ^ now.

jkbecker:
@fihzy, sorry, I made some changes on my end before looking at your comment. Where is this code snippet from? Is it from `get_running_archive_jobs? The `rsync_dest()` function will actually ignore the second argument completely, so it doesn't really matter what goes in there. Or am I looking in the wrong spot?

fihzy:
> @fihzy, sorry, I made some changes on my end before looking at your comment. Where is this code snippet from? Is it from `get_running_archive_jobs`? The `rsync_dest()` function will actually ignore the second argument completely, so it doesn't really matter what goes in there. Or am I looking in the wrong spot?

Yeah, it's from get_running_archive_jobs(). I do pass in that 2nd argument, and the originally submitted patch, rsync_dest() is already updated to use the second argument for mode=remote. So that original code and my tiny path makes the local archiving work on my box that is doing double duty as plotter + storage for another node that is rsync'ing to it. So satisfied with this software, thanks for all your work on it. I wish I had a better grip on git to interact and submit a patch so as not to make this harder :-)

altendky:
Development has continued in https://github.com/ericaltendorf/plotman/pull/627 building on this base.  Thanks for getting it started...  `:]`

PR #89, opened 2021-04-07
Make analyze subcommand work

jkbecker:
This should be fixed ASAP because for me the `analyze` command just didn't work any more.
Also, the analyze method is actually just static, which simplifies the call to a one-liner.

### DIFF
--- a/src/plotman/analyzer.py
+++ b/src/plotman/analyzer.py
@@ -8,120 +8,118 @@ import texttable as tt
 from plotman import plot_util
 
 
-class LogAnalyzer:
-    @staticmethod
-    def analyze(logfilenames, bytmp, bybitfield):
-        data = {}
-        for logfilename in logfilenames:
-            with open(logfilename, 'r') as f:
-                # Record of slicing and data associated with the slice
-                sl = 'x'         # Slice key
-                phase_time = {}  # Map from phase index to time
-                n_sorts = 0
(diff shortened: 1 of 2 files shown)


src/plotman/analyzer.py
```diff
@@ -9,7 +9,8 @@
 
 
 class LogAnalyzer:
-    def analyze(self, logfilenames, bytmp, bybitfield):
+    @staticmethod
+    def analyze(logfilenames, bytmp, bybitfield):
```
altendky, line 13: I personally just move stuff like this out as a freestanding function since it has no hard ties to the class.  Do you have any strong feelings for not doing that?

jkbecker: Seems like a simple method would work just fine - I wasn't sure if there was any backstory to this standalone class (plans to refactor it to something more OOP, etc.) so I didn't want to touch the external interface.

ericaltendorf: No particular reason for this to be a class method right now, no, it would be fine to move it out.

altendky: Oh hey, it's even the only thing at all in the class.  `:]`  Yeah, change requested here...

altendky (changes requested):

altendky (approved):

PR #90, opened 2021-04-07
fix: include tests in the package

altendky:
Also test the installed tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,17 @@ jobs:
         task:
           - name: Test
             tox: test
+        include:
+          - task:
+              name: Check
+              tox: check
+            os:
+              name: Linux
+              runs-on: ubuntu-latest
+            python:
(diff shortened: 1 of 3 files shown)


altendky:
`check-manifest` checks to make sure your VCS files match what will be included in your sdist.  So, for example, if you miss a test directory like we did, it complains.  (albeit, this is a made up scenario to show it working)

```console
$ venv/bin/check-manifest --verbose .
listing source files under version control: 25 files and directories
building an sdist: plotman-0.1.1.tar.gz: 26 files and directories
copying source files to a temporary directory
building a clean sdist: plotman-0.1.1.tar.gz: 25 files and directories
lists of files in version control and sdist do not match!
missing from VCS:
  src/plotman/_tests/blue/test_x.py
```

jkbecker:
Looks sensible to me!

Disclaimer: I know absolutely nothing about tox. But what I'm seeing makes sense.

PR #92, opened 2021-04-07
skip_install = true for tox check environment

altendky:


altendky:
Hmm, I think this doesn't work since we use the extras to specify the check dependencies.  Oh well, it isn't that expensive to install plotman as well.

PR #93, opened 2021-04-07
fix: handle plot file times under non-english locales

### LINKED ISSUE #29: Crash on unexpected datetime format
ValueError: time data 'Fri Feb 12 16:02:56 2021' does not match format '%a %b  %d %H:%M:%S %Y'

altendky:
Fixes #29

Draft for:
- [x] Self review
- [x] Merge of https://github.com/ericaltendorf/plotman/pull/94 to drop 3.6 support so we always have `importlib.resources` instead of having to use `importlib_resources` for now.  Which isn't really a big deal...  I dunno.  Whatever.


altendky:
I think my biggest doubt is about the goodness of bothering to test the whole log file loading right now including the funky work-around for the overly-ambitious `.__init__()`.  Since I extracted the time parsing to a separate function, it may make sense to drop all the log file bits and just test that function directly instead.  Then again, it might be nice to have the log file there for building future tests on.

ericaltendorf:
I looked over this enough to mostly understand it, but I don't have enough experience to have an opinion if it's a great idea or not.  Rubber-stamping approval.

ericaltendorf (approved):

altendky:
The faux-job test class is not good.  Pendulum is well liked.  Its time objects inherit from stdlib ones so they are interchangeable, one way anyways.  We could have just temporarily set the locale but that's global process state and thus hazardous to threading.  Maybe we could decide that isn't an issue here.

PR #94, opened 2021-04-07
Drop Python 3.6

altendky:
@jkbecker wins.  There are various tidbits in 3.7 that will be nice to have around.  Specifically `importlib.resources` for the example log file in #93.


altendky:
Could just use `importlib_resources` but...  I dunno, whatever.

PR #99, opened 2021-04-08
fix: deal with missing iowait

altendky:
Helps with at least one case from https://github.com/ericaltendorf/plotman/issues/38.


altendky:
Showing a `-` or something would be nice instead of `-1s`.

altendky:
Shows `-` for @Handri-Kosada on macOS and `4s` here on Linux at the moment.

ericaltendorf (approved):

PR #105, opened 2021-04-09
feat: add option to skip terminal jobs in analyze

ericaltendorf:


ericaltendorf:
Looks like some conflicting changes happened while I was doing this.  Yay, now I get to figure out merge conflicts in git.

altendky:
At this point I mostly use pycharm for things that git can't resolve itself.  If not pycharm then kdiff3 or meld.  I have all three setup as git mergetools so I can `git mergetool --tool=charm`.  A lot of the time they can just resolve either everything or almost everything automatically.  But sure, probably also good to be able to do stuff with git directly.

jkbecker:
I'm fairly sure it's because #89 was merged and it changes the indent of the analyze function 😞

ericaltendorf:
OK, I think I made it through.  Ready for review @altendky !

src/plotman/analyzer.py
```diff
@@ -17,21 +17,30 @@ def analyze(logfilenames, bytmp, bybitfield):
...
                     n_sorts = 0
                     n_uniform = 0
 
+                    seq_num = int(m.group(1))
+                    seq_total = int(m.group(2))
+                    is_first_last = seq_num == 1 or seq_num == seq_total
+
+                # Temp dirs.  Sample log line:
+                # Starting plotting progress into temporary dirs: /mnt/tmp/01 and /mnt/tmp/a
+                m = re.search(r'^Starting plotting.*dirs: (.*) and (.*)', line)
```
altendky, line 42: Can't wait for this to get caught up on a path like `C:\Users\Jane and John\Desktop\Chia`.  `:]`  Not that I have a proposed change...  so don't worry about it.

src/plotman/analyzer.py
```diff
@@ -17,21 +17,30 @@ def analyze(logfilenames, bytmp, bybitfield):
...
+            is_first_last = False
 
             # Read the logfile, triggering various behaviors on various
             # regex matches.
             for line in f:
                 # Beginning of plot job.  We may encounter this multiple
                 # times, if a job was run with -n > 1.  Sample log line:
-                # Starting plotting progress into temporary dirs: /mnt/tmp/01 and /mnt/tmp/a
-                m = re.search(r'^Starting plotting.*dirs: (.*) and (.*)', line)
+                # 2021-04-08T13:33:43.542  chia.plotting.create_plots       : INFO     Starting plot 1/5
```
altendky, line 27: I guess there's been some log format changes?  Do you know what chia-blockchain version?

altendky (commented):
It reads ok.  I didn't run it.  Normal regex hazards are present, but that's life with logs meant for humans.  I'll approve pending consideration of the chia version dependence.

altendky (approved):

src/plotman/analyzer.py
ericaltendorf: Not quite, it's more a confusing diff.  Behavior is essentially the same as before (the previous regex can still be seen below, in the next matcher), we're just also matching on an additional earlier line in the log, and we've moved some of the data-initialization here because it's now the first line that we match.

src/plotman/analyzer.py
ericaltendorf: Yeah.........

PR #106, opened 2021-04-10
Find chia in env if needed

### LINKED ISSUE #80: Be able to find chia command without activating the env
I'm not really sure what the priority should be in the case where there's a `chia` on the `$PATH` and also in the same env as plotman, but at least in the case where `chia` is not on the on the `$PATH` and is available in the env, it seems like it should be used?  Or maybe it must be configured?  At a high level this relates to being able to run plotman without activating the environment.

`sysconfig.get_path()` can get us the scripts (bin) path.

```console
$ venv/bin/python -c 'import sysconfig; print(sysconfig.get_path("scripts"))'
/farm/venv/bin
```

Or, maybe we can get the executable path from `psutil` when there are existing plots and if they all match then we can presume that's the `chia` to use?

altendky:
Fixes #80.


altendky:
I did run a k25 the whole way through and got a plot out the other end.  I dunno, I do think plotman should be able to work without activating the environment but there are various aspects of this that make me a bit itchy.  Let me know which bits do that for you.

And yeah, I may yet drop PyPy depending if Chia works with it.  Which it probably doesn't.

ping: @ericaltendorf, @jkbecker, @chelseadole

src/plotman/manager.py
```diff
@@ -112,7 +160,8 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
             logfile = os.path.join(dir_cfg['log'],
                     datetime.now().strftime('%Y-%m-%d-%H:%M:%S.log'))
 
-            plot_args = ['chia', 'plots', 'create',
+            plot_args = [CHIA_EXECUTABLE.command, 'plots', 'create',
+                    '--override-k',
```
chelseadole, line 164: Should `--override-k` be here?

chelseadole (commented):
Pretty clever solution to a finnicky situation. As is often the case with these kind of issues, there's likely a lot of room for edge cases, but I'm not sure they're worth the time they'd take to suss out, as they're likely quite specific.

The override of the user's environment makes me a bit itchy, but to be fair it's only for the duration of the single command

src/plotman/manager.py
altendky: Heh, nope...  that was for testing.  Thanks.  (side note, we still need to add that to the config.  but that comes after #107)

```suggestion
```

Issue #84, opened 2021-04-04
config.yaml should not be searched based on the current directory.


jkbecker:
There should be a central location for config.yaml ([see this discussion](https://github.com/ericaltendorf/plotman/pull/61#issuecomment-812967363).

I'm making this issue mostly because @altendky suggested I split this feature off of #61.

ericaltendorf:
I'm not sure how you want to group things here, but it sounds like we want
- A default location (like ~/.plotman.yaml? ~/.plotman/config.yaml?)
- A way to create such a default config
- A way to override the config file location on the command line (useful if you're plotting as a service for others, or if you're benchmarking and doing testing with different config settings)

altendky:
Sounds right to me.

IIRC @jkbecker suggested appdirs.  I haven't used it but it is what I would reach for.  It would use `~/.config/plotman/` for config on Linux, for example, and the appropriate locations on other OS.

https://pypi.org/project/appdirs/

jkbecker:
Appdirs gives you good default locations for config files, as well as logfiles.
I would definitely implement it in a way that allows for the search path to be extended + overridden, i.e. by passing in optional arguments or setting an environment variable.
Also, plotman should maybe have a `plotman config` command which can print the logfile location to make it easier to find - that command should be sensitive to environment variables as well, so something like
```
> plotman config path
/home/chia/.config/plotman/config.yaml
> export PLOTMAN_CONFIG=/somewhere/else/plotman.yaml
> plotman config path
/somewhere/else/plotman.yaml
> plotman config info
Active: /somewhere/else/plotman.yaml (file exists ✔︎)
Inactive (overridden by $PLOTMAN_CONFIG):
    - Default: /home/chia/.config/plotman/config.yaml (file exists ✔︎)
 > plotman --config custom-config.yaml config path
/home/chia/custom-config.yaml
 > plotman --config custom-config.yaml config info
FILE                                    TYPE              EXISTS   SCHEMA      ACTIVE
/home/chia/custom-config.yaml           --config          Yes      ERROR       Active
/somewhere/else/plotman.yaml            $PLOTMAN_CONFIG   Yes      OK          Inactive (--config override)
/home/chia/.config/plotman/config.yaml  default           Yes      OK          Inactive (--config override)
``` 

Is this over-engineered?
I feel the same thing could be done for logfiles (appdirs also has defaults for that) - that would help a lot with the tons of support requests that result from a faulty logfile location in the config file.

[...]

PR #107, opened 2021-04-10
Validate config.yaml contents (#77)

chelseadole:
Addresses #77 

* Adds `desert`-based dataclasses to configuration.py to deserialize config.yaml contents
* Changes all references of dictionary bracket notation --> dataclass dot notation
* Adds tests

In the discussion for #77 , @jkbecker mentioned potentially making config-related changes discussed in #84 . I did not get to this, because the PR reached a critical length where I think it's safest to separate concerns. :)


chelseadole:
RE: failing tests -- tests seem to be failing because the `ci.yml` builds the `plotman` package using dist/wheel rather than cloning, so `config.yaml` is not present. 

Is this desired? Checking out the `setup.cfg`, I see:
```
[options.data_files]
config = config.yaml
```

but I don't see anything about `config.yaml` in the `MANIFEST.in`. What is the desired behavior RE: config.yaml being present when a user installs using pip? 

For these tests, it would be possible to mock out the `yaml.load()`, so that it's not actually necessary for `config.yaml` to be present. However, IMO that somewhat defeats the purpose of the main test -- as it would be helpful to check the validity of the default/example `config.yaml` in this repo without expecting contributors to know they have to update a test-only mirror when adding new config values. Thoughts?

jkbecker:
Awesome!

> RE: failing tests -- tests seem to be failing because the ci.yml builds the plotman package using dist/wheel rather than cloning, so config.yaml is not present.

Haha, no, this is not desired. But also not a surprise.

When I proposed #61, we had an at-length discussion there about the config file being treated as a separate issue because it's just an intermediary weirdness and existing users expect it to sit in their current working directory anyways. The actual issue for taking care of config is #84 and I have not gotten around to implementing it. In fact I haven't touched it at all - if you're in the zone and this seems like a good fit, feel free to look into it, ... otherwise maybe @altendky has a good idea of how to hotfix the CI for now (he introduced it, it's mostly black magic to me) to make this one work in isolation?

src/plotman/archive.py
```diff
@@ -90,15 +90,14 @@ def archive(dir_cfg, all_jobs):
     contention on the plotting dstdir drives.  Returns either (False, <reason>) 
     if we should not execute an archive job or (True, <cmd>) with the archive
     command if we should.'''
-
-    dstdirs = dir_cfg['dst']
-    arch_cfg = dir_cfg['archive']
```
altendky, line 95: Exactly the sort of stuff I wanted gone...  `:]`

src/plotman/archive.py
```diff
@@ -117,26 +116,25 @@ def archive(dir_cfg, all_jobs):
...
     if not archdir_freebytes:
         return(False, 'No free archive dirs found.')
     
     archdir = ''
     available = [(d, space) for (d, space) in archdir_freebytes.items() if 
                  space > 1.2 * plot_util.get_k32_plotsize()]
-    if len(available):
-        index = arch_cfg['index'] if 'index' in arch_cfg else 0
-        index = min(index, len(available) - 1)
+    if available:
```
altendky, line 126: I'm not requesting a change here, but I would have gone the other way if anything.  `if len(available) > 0:`  In this case it doesn't really matter since it is necessarily a list and that is forced on the line above.  But, I prefer explicitness in tests.  There was a case recently in plotman where a 0 was misprocessed the same as a `None`, IIRC.  Anyways, opinions etc.  Just sharing mine so we can all learn each other's tendencies.  `:]`

src/plotman/interactive.py
```diff
@@ -149,7 +143,7 @@ def curses_main(stdscr):
                                     stderr=subprocess.STDOUT,
                                     start_new_session=True)
 
-                archdir_freebytes = archive.get_archdir_freebytes(dir_cfg['archive'])
+                archdir_freebytes = archive.get_archdir_freebytes(cfg.directories)
```
altendky, line 146:
```suggestion
                archdir_freebytes = archive.get_archdir_freebytes(cfg.directories.archive)
```

src/plotman/interactive.py
```diff
@@ -179,21 +174,21 @@ def curses_main(stdscr):
         #
 
         # Directory prefixes, for abbreviation
-        tmp_prefix = os.path.commonpath(dir_cfg['tmp'])
-        dst_prefix = os.path.commonpath(dir_cfg['dst'])
+        tmp_prefix = os.path.commonpath(cfg.directories.dst)
```
altendky, line 177: These sorts of changes are mind numbing.  I get it.  `:]`
```suggestion
        tmp_prefix = os.path.commonpath(cfg.directories.tmp)
```

src/plotman/manager.py
```diff
@@ -50,44 +50,36 @@ def phases_permit_new_job(phases, d, sched_cfg, dir_cfg):
...
     youngest_job_age = min(jobs, key=job.Job.get_time_wall).get_time_wall() if jobs else MAX_AGE
-    global_stagger = int(sched_cfg['global_stagger_m'] * MIN)
+    global_stagger = int(sched_cfg.global_stagger_m * MIN)
     if (youngest_job_age < global_stagger):
-        wait_reason = 'stagger (%ds/%ds)' % (
-                youngest_job_age, global_stagger)
-    elif ('global_max_jobs' in sched_cfg and
-            len(jobs) >= sched_cfg['global_max_jobs']):
-        wait_reason = 'max jobs (%d)' % sched_cfg['global_max_jobs']
+        wait_reason = 'stagger (%ds/%ds)' % (youngest_job_age, global_stagger)
```
altendky, line 78: No need to undo them, but there are a few places where there have been unrelated changes.  I do find it is easier to handle diffs without that.  As to formatting, I vaguely 'expect' to get black in place at some point.  Though, that's just me.  I haven't gotten any team consensus on it.

src/plotman/manager.py
```diff
@@ -101,35 +93,36 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
...
+            if plotting_cfg.farmer_pk is not None:
                 plot_args.append('-f')
-                plot_args.append(plotting_cfg['farmer_pk'])
-            if 'pool_pk' in plotting_cfg and plotting_cfg['pool_pk']:
+                plot_args.append(plotting_cfg.farmer_pk)
+            if plotting_cfg.pool_pk is not None:
                 plot_args.append('-p')
-                plot_args.append(plotting_cfg['pool_pk'])
-            if 'tmp2' in dir_cfg:
+                plot_args.append(plotting_cfg.pook_pk)
```
altendky, line 122:
```suggestion
                plot_args.append(plotting_cfg.pool_pk)
```

altendky (changes requested):
Several specific comments and some general questions, but overall this is having exactly the effect I'd hoped for.  Simple and complete definition of the configuration along with explicit errors for users rather than random code giving `TypeError` or such.  `:]`  Thanks for all the work on this.  I have a few followup changes that I want to do that will be much more sensible because of this improvement.

1) For all (@jkbecker, @ericaltendorf), do we want to use `attrs` or `dataclasses`?  https://www.attrs.org/en/stable/why.html#data-classes gives a brief explanation.  I default to attrs.  It shouldn't make much of a difference to the PR.

2) Yeah, we need to decide how to handle the config file.  I haven't really made out a whole plan, but I'll write some stuff and see where we end up.  We want to be able to, after installation, write out a file to a given location with a 'default' configuration.  The content could be sourced from an existing file or from a function that creates instances of the new configuration classes.  If it is a file, then yes, it should be validated in the tests as @chelseadole pointed out.  I suppose it is pretty handy to have a file to link to in the repository and talk about.  Maybe that's enough to make up my mind on using a file.  I don't super care where it lives but I might default to something like `src/plotman/resources/config.yaml` (or `plotman.yaml`).  https://github.com/ericaltendorf/plotman/pull/93/files provides a reference for including data files and using `importlib.resources`.  Tests can get at the data and the runtime can get to it as well to write out the default config for the user.

I expect the `options.data_files` will go away.  I think `MANIFEST.in` and `include_package_data = True` in `setup.cfg` will cover the data files we need.

Thanks again for making this big improvement here.  The list of comments looks unrepresentatively long `:[` but I think figuring out the config file location etc might be the biggest thing, and it shouldn't be too bad...  `:]`

src/plotman/interactive.py
chelseadole: Ahh sorry! Thanks for catching

src/plotman/configuration.py
```diff
@@ -1,9 +1,69 @@
...
+    rsyncd_user: str
+    index: Optional[int] = 0  # If not explicit, "index" will default to 0
+
+@dataclass
+class Directories:
+    log: str
+    tmp: List[str]
+    dst: List[str]
+    tmp2: Optional[str] = None
+    tmp_overrides: Optional[Dict[str, Dict[str, int]]] = None
```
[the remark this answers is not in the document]
chelseadole: From my understanding, names of the keys in `tmp_overrides` are variable - as they should be the names of configurable directories. In the example config.yaml, they are `/mnt/tmp/00`, etc. 

From my understanding, `dataclasses` need consistent field naming

src/plotman/configuration.py
```diff
@@ -1,9 +1,69 @@
...
+# Data models used to deserializing/formatting config.yaml files.
+
+@dataclass
+class Archive:
+    rsyncd_module: str
+    rsyncd_path: str
+    rsyncd_bwlimit: int
+    rsyncd_host: str
+    rsyncd_user: str
+    index: Optional[int] = 0  # If not explicit, "index" will default to 0
```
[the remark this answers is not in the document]
chelseadole: This makes sense. Updated to remove `Optional`, and did the same to `tmpdir_stagger_phase_limit`, which has the same problem

src/plotman/archive.py
chelseadole: Given that `available` is produced via a list comprehension, it's safe from `None` issues, but I'm totally happy to change to `if len(available) > 0`. My preference is either my original code or what you suggested instead of `if len(available)` :)

src/plotman/configuration.py
```diff
@@ -1,9 +1,69 @@
...
+
+import desert
 import yaml
 
 
-def get_path():
-    return 'config.yaml'
+def get_validated_configs():
+    """Return a validated instance of the PlotmanConfig dataclass with data from config.yaml."""
+    config_path = "config.yaml"
```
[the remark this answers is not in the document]
chelseadole: Ok! I changed because once I added `get_validated_configs()`, `get_path()` wasn't being utilized anymore. I will change back given #84 👍

src/plotman/manager.py
chelseadole: Noted! I'm personally a fan of `black` purely because it reduces developer infighting time RE: stylistic change 😅

chelseadole:
@altendky @jkbecker 

Thank you both for the review, comments, and context. To unblock myself in this PR, I will change `config.yaml` to be `src/plotman/resources/plotman.yaml` (I like the rename suggestion to `plotman.yaml`), and will add to `MANIFEST`, etc so it's packaged correctly. 

> We want to be able to, after installation, write out a file to a given location with a 'default' configuration. 

My personal approach would be adding new args similar to what's suggested in #84, so when you `pip install` plotman, you're able to run something like: `plotman config generate`, which would create a config file at a pre-set location on your computer based on appdirs. Assuming we include a default `plotman.yaml` in the package as discussed above, we can just copy that file over. With that set up, we'd be able to run `plotman config path`, etc as suggested by @jkbecker [here](https://github.com/ericaltendorf/plotman/issues/84#issuecomment-815904053)

There's a lot more detail discussed in #84, and I don't want to overcrowd this PR. How about I do the above work to unblock myself with testing, and leave additional commandline features with `plotman config <...>` to a future PRs?

src/plotman/configuration.py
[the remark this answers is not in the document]
altendky: Agreed, I was referring to the inner not the outer.  Let's make sure I'm getting my story straight here.

```yaml
directories:
        tmp_overrides:
                "/mnt/tmp/00":
                        tmpdir_max_jobs: 5
```

So something like this with another class defined for the override information.  Admittedly with just one attribute presently.

```python3
tmp_overrides: Optional[Dict[str, TemporaryDiskOverride]] = None
```

The outer dict stays with the keys being whatever temporary directory paths people specify.  This may just work or may require some tricks for desert/marshmallow.  I'm not certain.

altendky:
Agreed that the various discussions about finding the config file should be addressed in a separate PR.

src/plotman/configuration.py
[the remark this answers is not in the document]
chelseadole: Changed 👍

chelseadole:
@altendky Let me know what you think of the added changes. It now supports:

```bash
$ plotman config show
$ plotman config generate
```

For now, it only performs `generate` with a default location (no configurable location/path) but that could be a future improvement. I also added the default `resources/plotman.yaml` file to the package as discussed above

README.md
```diff
@@ -166,7 +166,7 @@ mode or the command line mode.
 
 There are many bugs and TODOs.
 
-Plotman will always look in the current directory for the `config.yaml` file
+Plotman will always look in the current directory for the `plotman.yaml` file
```
jkbecker, line 169: Is this still correct or is that outdated, now that `generate` actually drops the config file into the config directory defined by `appdirs`?

README.md
```diff
@@ -179,9 +179,9 @@ Installation for Linux:
 2. Then, install Plotman using the following command:
 
        pip install --force-reinstall git+https://github.com/ericaltendorf/plotman@development
-3. Plotman will look for `config.yaml` in your current working directory when you run it.
-   You can find [an example configuration file](./config.yaml) in the project source that can be used
-   as a starting point.
+3. Plotman will look for `plotman.yaml` in your current working directory when you run it.
```
jkbecker, line 182: ditto re: appdirs-defined config file location as generated per `generate`

jkbecker (commented):
Are these two sections still correct? They seem to be superseded by what the `generate` command does?

jkbecker:
This looks awesome! I feel like it's probably clean to stick with just one default config location in this PR, and address the overloadable config file situation mentioned in #84 in a separate follow-up PR (I'm happy to work on that one but I'm currently super maxed out with "work" work, so... not mad if you get to tackling it first - we can play by ear.)

README.md
chelseadole: I updated the README to be accurate. Thanks for the catch!

chelseadole:
> This looks awesome! I feel like it's probably clean to stick with just one default config location in this PR, and address the overloadable config file situation mentioned in #84 in a separate follow-up PR (I'm happy to work on that one but I'm currently super maxed out with "work" work, so... not mad if you get to tackling it first - we can play by ear.)

Yes, I wanted to start with a constant config location before adding complexity. I'm also ramping up on a new project at "work" work, so I'll see -- but if I have time I'll try to tackle it!

altendky (changes requested):
Thanks for keeping at this.  `:]`

src/plotman/plotman.py
```diff
@@ -95,45 +97,73 @@ def main():
...
+                    overwrite = input(
+                        f"A 'plotman.yaml' file already exists at the default location: '{config_file_path}' \n\n"
+                        "\tInput 'y' to overwrite existing file, or 'n' to exit without overwrite."
+                      ).lower()
+                    if overwrite == 'n':
+                        return "\nExited without overrwriting file"
+
+            # Copy the default plotman.yaml (packaged in plotman/resources/) to the user's config file path,
+            # creating the parent plotman/ directory if it does not yet exist
+            config_directory_path = configuration.get_directory_path()
```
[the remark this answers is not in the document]
chelseadole: Good point, changed

setup.cfg
```diff
@@ -52,10 +55,11 @@ dev =
 test = 
     check-manifest
     pytest
+    pytest-mock
     pyfakefs
 
 [options.data_files]
-config = config.yaml
+config = src/plotman/resources/plotman.yaml
```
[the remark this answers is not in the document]
chelseadole: I tried removing this in an earlier iteration, and ci failed without it. For the sake of moving onwards, could we keep "package setup/MANIFEST cleanup" as a future TODO, or would you like me to mess around with it a bit more here?

src/plotman/plotman.py
```diff
@@ -26,20 +24,24 @@ def parse_args(self):
...
+        sp.add_parser('plot', help='run plotting loop')
 
-        p_archive = sp.add_parser('archive',
-                help='move completed plots to farming location')
+        sp.add_parser('archive', help='move completed plots to farming location')
+
+        p_config = sp.add_parser('config', help='display or generate plotman.yaml configuration')
+        sp_config = p_config.add_subparsers(dest='config_subcommand')
+        sp_config.add_parser('generate', help='generate a default plotman.yaml file and print path')
+        sp_config.add_parser('show', help='show path to current plotman.yaml file')
```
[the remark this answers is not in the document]
chelseadole: I like `path`, but personally prefer `generate` :) Changed

setup.cfg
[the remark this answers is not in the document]
altendky: That's fine.  There's another file here in the list to get as well so making that a dedicated PR sounds good.  Thanks for trying it out.

altendky (approved):
Thanks @chelseadole, definitely glad to see this change.  Sorry for the wait.

Issue #96, opened 2021-04-07
`plotman archive` does not, in fact, archive.


jkbecker:
Instead, it just returns a tuple with the command list that could perform an archive command.

https://github.com/ericaltendorf/plotman/blob/349d13ab315518d22d55b04ceead17b3addb4ff7/plotman.py#L135

https://github.com/ericaltendorf/plotman/blob/349d13ab315518d22d55b04ceead17b3addb4ff7/archive.py#L138

cmeyer90:
I attempted a fix for this - see PR here: https://github.com/ericaltendorf/plotman/pull/109

Still testing but appears to work fine in 'plotman archive' mode.

chadmccune:
I'm experiencing this problem in 0.3.1...

in 'plotman interactive', it archives, but in 'plotman archive', it just loops never executing rsync as far as I can tell

altendky:
@chadmccune, that is a separate issue.  Please provide example outputs etc in a new ticket.

PR #109, opened 2021-04-13
Fix standalone archive and add rsync status

cmeyer90:


src/plotman/archive.py
```diff
@@ -130,13 +140,41 @@ def archive(dir_cfg, all_jobs):
         (archdir, freespace) = sorted(available)[index]
 
     if not archdir:
-        return(False, 'No archive directories found with enough free space')
-    
+        return 'No archive directories found with enough free space'
+
+    arch_jobs = get_running_archive_jobs(dir_cfg['archive'])
```
chelseadole, line 145: Can you pull from `development` on this branch? There have been changes to how we reference configs internally, so references to `dir_cfg["archive"]` should all use dot notation now: `dir_cfg.archive`

src/plotman/archive.py
```diff
@@ -130,13 +140,41 @@ def archive(dir_cfg, all_jobs):
...
+    if arch_jobs:
+        # Build status list
+        status_list = 'Cannot retrieve status.'
```
chelseadole, line 148: Calling this `status_list` with type str seems misleading. Can you either make it a list, or rename?

src/plotman/archive.py
```diff
@@ -130,13 +140,41 @@ def archive(dir_cfg, all_jobs):
...
+        for j in jobs:
+            status_list = ''
+            try:
+                line = j[0].get_nowait() # or q.get(timeout=.1)
+            except Empty:
+                line = 'No data yet.'
+            if (line is None) or (line in j[1]):
+                line = 'No data yet.'
+            status_list += 'Moving ' + j[1] + ' to ' + j[2] + '. Progress: ' + line + '\n'
```
chelseadole, line 157: Why reassign `status_list` to empty string at the beginning of each for loop when you're trying to append to it with new status every time?

cmeyer90: Thank you for the feedback! I have made some untested changes, will resume tomorrow and test when I have a plot available then perhaps we can merge.

chelseadole: Sounds good! And thank you for tackling 😁

chelseadole (approved):
Looks good to me. @altendky do you have any thoughts?

src/plotman/archive.py
```diff
@@ -15,6 +17,13 @@
 
 # TODO : write-protect and delete-protect archived plots
 
+jobs = []
```
ericaltendorf, line 20: are these also archive jobs?  it's unclear...  in the rest of the code base, "jobs" on its own always refers to plotting jobs.

cmeyer90: will make this more descriptive. doing some testing right now.

cmeyer90:
I did some more testing on the changes this morning. Everything is functional with `plotman archive` and `plotman interactive` but there seems to be an occasional issue with the rsync output showing inaccurate data for file transfer times and percentages. Though the transfer speed always appears correct, every so often everything else seems to have trouble updating, especially on faster networks.

I was hoping I could track down what causes this but I think it is a bug with rsync or more likely with the way rsync's --outbuf=L option works when streaming to a pipe. I still find the output very useful as I can see the byte count increase and transfer speed itself. Screenshot below for example over 10 gig ethernet:

Screen Shot 2021-04-18 at 10 26 02 AM

And when things are working well, from within `plotman interactive`:
(This is on a smaller screen but the number on the far right is the transfer speed over 1 gig ethernet)
Screen Shot 2021-04-18 at 10 27 55 AM

@altendky @chelseadole

chelseadole:
> @ericaltendorf @chelseadole

Thanks for the screenshots/notes. With the assumption that it's an rsync issue (is there anything you've found online to support that theory?) and that the issue is occasional bad _reporting_, not breaking of the rsync functionality, I'd be happy with this current version. 

If rsync's internal implementation really is the core issue, it doesn't seem like a great use of your time to hunt that down IMHO

cmeyer90:
> > @ericaltendorf @chelseadole
> 
> Thanks for the screenshots/notes. With the assumption that it's an rsync issue (is there anything you've found online to support that theory?) and that the issue is occasional bad _reporting_, not breaking of the rsync functionality, I'd be happy with this current version.
> 
> If rsync's internal implementation really is the core issue, it doesn't seem like a great use of your time to hunt that down IMHO

I fixed it - not an rsync bug but rather an implementation error from my side. The thread that handles the logger was using a queue and therefore pulling the oldest entry from a long queue of entries every 60 seconds.

I have switched to a stack so the latest line of data is retrieved from rsync. Additionally I have limited the stack size to 3 entries in an attempt to avoid any possible memory leaks.

This is not thoroughly tested, but I just did a transfer with `plotman archive` and the output is much more useful:

Screen Shot 2021-04-18 at 11 00 11 PM

The only thing missing is a 'transfer complete' type of confirmation, but generally you can infer that a transfer has completed when it stops generating logs.

@chelseadole @ericaltendorf - please feel free to install this branch if you haven't already and give it a shot!

altendky:
Hi @cmeyer90, thanks for the work here.  We definitely do need to get the archive command fixed.  rsync progress sounds nice too.  I do have some concerns though.  To get each piece through I think it would be good to have separate PRs for them.

I think that fixing the archive mode could be done without changing the present design where `archive()` returns the command line to run instead of running it inside of the `archive()` function.  Apparently we don't presently have any tests, but returning the command is easier to write tests for than if the function actually tries to run the subprocess.  Would you be interested in submitting a single purpose PR that just fixes the `plotman archive` subcommand?

On the rsync process side there's a bit more to it.  I didn't look into the details, but at a high level I see that we are able to monitor plotting processes without threads and without state.  The latter means that the user can quit plotman and reopen it and still have the same complete functionality.  This won't work when piping the rsync output to the plotman process.  I hate to turn away work like this but I do find that the stateless aspect of plotman is pretty nice and would like to retain it as features are added.  I would be happy to work with you to try to adapt this effort to fit more closely with the existing plotman design.

Thanks again for your work here.  We need all the help we can get.  I'm sorry that I'm asking for so many changes.  I hope we can get through them and get your fix and feature merged.

Side note, I have been using `progress --monitor` for keeping track of copies and moves and rsyncs.  It isn't integrated into plotman so it isn't really an alternative, but it may none the less be of interest to some people that read this.

cmeyer90:
> Hi @cmeyer90, thanks for the work here. We definitely do need to get the archive command fixed. rsync progress sounds nice too. I do have some concerns though. To get each piece through I think it would be good to have separate PRs for them.
> 
> I think that fixing the archive mode could be done without changing the present design where `archive()` returns the command line to run instead of running it inside of the `archive()` function. Apparently we don't presently have any tests, but returning the command is easier to write tests for than if the function actually tries to run the subprocess. Would you be interested in submitting a single purpose PR that just fixes the `plotman archive` subcommand?
> 
> On the rsync process side there's a bit more to it. I didn't look into the details, but at a high level I see that we are able to monitor plotting processes without threads and without state. The latter means that the user can quit plotman and reopen it and still have the same complete functionality. This won't work when piping the rsync output to the plotman process. I hate to turn away work like this but I do find that the stateless aspect of plotman is pretty nice and would like to retain it as features are added. I would be happy to work with you to try to adapt this effort to fit more closely with the existing plotman design.
> 
> Thanks again for your work here. We need all the help we can get. I'm sorry that I'm asking for so many changes. I hope we can get through them and get your fix and feature merged.
> 
> Side note, I have been using `progress --monitor` for keeping track of copies and moves and rsyncs. It isn't integrated into plotman so it isn't really an alternative, but it may none the less be of interest to some people that read this.

@altendky All great feedback! It's nice to have some guidance here and you make some valid points. These are things I would love to collaborate with you on.
 
As a start, per your suggestion, I will create another branch that implements `plotman archive` without bundling the subprocess call within the archive() function. In briefly reviewing the project structure, I think a good goal might be to move some of the code that would be shared between plotman.py and interactive.py (for calling the subprocess) into a 'common' folder. This is similar to the original method of moving everything into archive.py but more closely follows the original design of plotman. My goal is to avoid accidentally archiving with different settings between `plotman interactive` and `plotman archive`. I will think about this some more and welcome your feedback or any solutions you might have in your mind.

For the rsync piece - I introduced threading because reading the stdout for the rsync call seemed to block plotman from proceeding with other tasks until the rsync data was returned. I tried to do this in a way that keeps with the stateless behavior that plotman originally implemented but it is entirely possible I overlooked something here. To add some depth, here is some background on how this PR works in its current state, as tested:
 - after quitting plotman, rsync jobs continue to run in the background. verified through htop.
 - upon resuming `plotman archive` or `plotman interactive` the jobs are still found using the original logic, but they no longer have knowledge of rsync output and instead return 'Arch job(s) already running. Cannot retrieve status'
 - the logging thread dies when rsync stops providing stdout. it terminates when the rsync job exits (no more stdout) or plotman exits (the user quit)

I am not sure if this meets the stateless requirements. I am happy to work together on this piece as it will need some tweaking regardless.

Let me start on #96 and we can move forward together from there!

altendky:
I appreciate your interest in keeping the duplication of code low.  I think we can achieve both goals.  Certainly at some point we'll be calling `Popen` or similar and that'll just have to be dealt with when we get around to testing that code.  Putting that code in a common function is fine by me as long as it does little else so that as much code as possible is just functions.  Alternatively, without having thought through how this would work out here, sometimes you can just return a dict of function parameters.  Such as `do_it, params = archive(etc)` and then `if do_it: Popen(**params)`.  Not having thought through the details yet I'm not fully suggesting that, just sharing another option that is nice at times.

On the other side, the solution with the plotting processes was to always launch them with output directed to the log files.  We can then read the log files at our leisure, including from subsequent invocations of plotman.  I haven't looked through the code closely enough to decide if I like the details, but the general approach seems useful.  The main downside is that you pile up logfiles.  But, given that the entire goal of this program is to pile up terabytes of data...  a few more log files probably isn't too bad.  `:]`  Do you think this could work with rsync as well?

cmeyer90:
Sounds good on item number one. I like your idea of using **params to unpack the result from archive(). Let me meditate on this some more and I will come up with a good solution.

On item number two - thank you for the clarification this makes more sense now. Please correct me if I am wrong but it sounds like the functionality implemented in this PR for rsync is ok, but we we still need to, in a stateless way, redirect the output of rsync to our plotman log files. I noticed that rsync does have a `--log-file=FILE` option which could do what we want, but will need to investigate further. It may even be good to separate the archival logs from the plotting logs. 

As for log files themselves, I have quite a lot of them so far! I think you are right that a few extra log files don't hurt when we are dealing in the terabytes, but maybe some type of log rotation system is in order. I have a shell script right now that handles some similar functions for me on reboot (clean up .tmp files, move logs to a different folder). It would be neat to have a flag for plotman that manages this - `plotman interactive -clean` for example.

altendky:
Yes, I can see plotman growing some log management bits.  Separate PR please.  `:]`

As to the overall functionality, I'll admit I'm a bit on the fence.  I'm not sure where the line is between overinvesting in this feature and providing useful tooling for the users.  For example, at what point should someone just use https://github.com/rjsears/chia_plot_manager instead?  There has been discussion of increased flexibility of the archiving mechanism which will make all the bits some amount more complicated.  At least maybe.  So on the one hand the rsync status showing in plotman sounds obviously great and on the other I'm not sure I see the big picture of where archiving in plotman is going.  But yeah, if we get the feature implemented inline with the present architecture then I expect to appreciatively merge it.  `:]`

Thanks for working through this.

PR #114, opened 2021-04-14
Handle plots create default arguments

altendky:
Draft for:
- [ ] Self testing
- [ ] Maybe just throw this out and figure out proper use of chia's own parser.  Though, there hasn't been official agreement on making a hard expectation that there is a chia installed in the same env as plotman and that is the one in use.  Though not requiring that kind of just leaves us hoping that we're matching.


altendky:
If we decide against https://github.com/ericaltendorf/plotman/pull/116 then we can reconsider this.

PR #115, opened 2021-04-16
Remove dead job.Job.check_status()

altendky:


jkbecker (approved):
Yup, seems orphaned to me.

PR #116, opened 2021-04-16
Use chia cli parser

altendky:
Draft for:
- [x] Handle the `--help` case.
  - https://replit.com/@altendky/TurbulentFelineDaemon-2#main.py
- [x] Self reflection over night.
- [x] Handle the `-h` case.
- [x] Drop PyPy testing in https://github.com/ericaltendorf/plotman/pull/125
- [x] Probably copy in and license acknowledge.  With some decisions about how to inspect we could then handle multiple chia versions.


ericaltendorf:
This is great!

altendky:
Note that this is still hand waving in terms of the subcommand parsing.  There could be arguments interleaved for parent subcommands in the command line without making it so that it isn't a plot command (present code would ignore such a process).  There may be more but at at least `venv/bin/chia --root-path my_path_to_chia_stuff plots create` should be valid.  But, I think it's fair to leave that for 'maybe later'.  Also, the choice of `-h` and `--help` is hard coded since `command.get_help_option_names(ctx=context)` only returns `--help` (comment left in the relevant code).  I didn't dig hard into the details of that.

chelseadole (approved):
Looks good to me :)

altendky:
This got reworked to copy in the Chia command parsing code rather than having the dependency on the chia-blockchain package.  In various cases people are installing chia without the Python package such as .dmg on macOS.  This approach also allows us to store multiple versions and in the future we could attempt to identify the version of the process and pick the proper parsing code for it.  Anyways...  maybe this is ready for real review and merging so we can move on from all these complaints...

ericaltendorf (commented):
It seems unfortunate that we need to copy chia code over, but if that's what seems like the best option, let's go for it.  Should it be documented somewhere what's going on?  I.e., which code came from chia and what the process should be for updating it / keeping it in sync?

src/plotman/chia.py
```diff
@@ -0,0 +1,83 @@
...
+        return max(self.by_version.items())[1]
+
+
+commands = Commands()
+
+
+@commands.register(version=(1, 1, 2))
+@click.command()
+# https://github.com/Chia-Network/chia-blockchain/blob/v1.1.2/LICENSE
+# https://github.com/Chia-Network/chia-blockchain/blob/v1.1.2/chia/cmds/plots.py#L39-L83
```
altendky, line 36: Here is the present indication of which code came from `chia-blockchain` and where it came from.  It does mark the specific lines copied in the link, though there could be a 'start of copy' and 'end of copy' mark in our code to be more explicit, if you want.  There should be no need to 'update' this specific code.  It is exactly for `chia-blockchain==1.1.2` and that "should" never change.  But, we will want to add newer (and possibly older) versions as well.  I mean...  there's already 1.1.3 and this hasn't even been merged.

So...  Yeah, I guess some maintenance documentation could be good.  I'll make a commit for adding 1.1.3 and some commentary somewhere can link to that.

On the one hand it is too bad that I'm going with copying the code.  But, the big flexibility it brings is the ability to have multiple versions available.  There are too many ways for programs to be distributed to bank on 'plotman should use the chia in its own env' unless we just mandate everyone do that.  But if anything, we probably ought to be going the other way and distributing plotman as a regular old program instead of exclusively as a Python package.  (this isn't a big deal to do, at least in a preliminary manner)  One of the issues that has come up a few times recently is macOS users that are running chia from the `.dmg`.  So, all in all this isn't that filthy.  Using the chia in the env wouldn't necessarily get us the proper parsing code for whatever random process we are looking at anyways.  So...  meh.  I think this isn't exactly the same as other cases of just copying code.

altendky:
Alrighty, there's some documentation.  Let me know if it is clear.

ericaltendorf:
Makes sense, thanks!

ericaltendorf (approved):

PR #117, opened 2021-04-17
yield configuration path in configuration path fixture

altendky:
https://docs.python.org/3.9/library/importlib.html#importlib.resources.path

If needed, importlib.resources.path() will write the resources to a temporary location.  This will be cleaned up after exiting the context manager.  Using yield leaves the fixture 'in' the context manager until the fixture is cleaned up.


jkbecker (approved):
OK

PR #118, opened 2021-04-17
change farmer_pk and pool_pk to type str

cciollaro:
currently when supplying these values you get the following exception:
```
marshmallow.exceptions.ValidationError: {'plotting': {'farmer_pk': ['Not a valid integer.'], 'pool_pk': ['Not a valid integer.']}}
```


altendky (approved):
I feel like these should maybe be something more specific so they can't be strings like `'the blue & green 37s are *$*#&'`, but `str` is still better than `int`.  Thanks for the fix @cciollaro.

PR #120, opened 2021-04-17
fix: printing and return code for config subcommands

altendky:


jkbecker (approved):
Yeah, this does look cleaner to me.

PR #132, opened 2021-04-19
Log instances of delays due to resource limitations and their causes

### LINKED ISSUE #129: Log instances of delays due to resource limitations and their individual causes
If a delay occurs that is in excess of the default stagger, record the duration of the added delay and the cause.  This would be useful for adjusting hardware in a plotter to maximize throughput, or for suggesting changes to make to more efficiently utilize the current hardware.  For example, if the config file has a stagger of 30 minutes, a user should expect 48 plots per day on average.  Any delays in the start of a plot would lower that number.  Knowing that a plot was delayed for an hour due to no eligible tmp directories would indicate that additional tmp drives are necessary, as opposed to more plotting instances.

Example:
Delayed 20 minutes - No eligible tmp drives
Delayed 35 minutes - Max jobs

13thProgression:
Fixes #129 
Include job age any time that a plot is held off
Insert the last delay message into the log if it isn't a standard stagger delay


altendky (commented):
@ericaltendorf, looks ok to me.  Any concerns adding the stagger info to the other statuses?

src/plotman/interactive.py
```diff
@@ -115,10 +116,16 @@ def curses_main(stdscr):
...
+                    if aging_reason is not None:
+                        log.log(aging_reason)
+                        aging_reason = None
                     log.log(msg)
                     plotting_status = '<just started job>'
                     jobs = Job.get_running_jobs(cfg.directories.log, cached_jobs=jobs)
                 else:
+                    # If a plot is delayed for any reason other than stagger, log it
+                    if msg.find("stagger") < 0:
+                        aging_reason = msg
```
ericaltendorf, line 128: So ... when we start a new plot job, we'll log the reason it had been delayed?  That could be useful.

ericaltendorf (approved):

src/plotman/interactive.py
13thProgression: It has helped me with tuning my machines by suggesting what resources are lacking.  The only downside that I have encountered is that the Log: count message no longer reflects the total number of jobs launched by plotman since delay messages are also counted.

altendky:
@13thProgression, thanks for joining the ranks of plotman developers.  Cheers.

PR #137, opened 2021-04-20
ci: also build on pushes to main

altendky:


jkbecker (approved):
Yeah, this makes sense. We can still discuss what exactly our workflow should be but can't hurt to make sure the CI runs against either.

PR #139, opened 2021-04-20
update default config.yaml values

cciollaro:
obviously we can't provide a config.yaml that works for everyone, but at this point bitfield is going to be better for the vast majority of people and the recommended max memory value (from the changelog at least) is 3389

### DIFF
--- a/src/plotman/resources/plotman.yaml
+++ b/src/plotman/resources/plotman.yaml
@@ -110,9 +110,9 @@ scheduling:
 plotting:
         k: 32
         e: False             # Use -e plotting option
-        n_threads: 8         # Threads per job
+        n_threads: 2         # Threads per job
         n_buckets: 128       # Number of buckets to split data into
-        job_buffer: 3389     # Per job memory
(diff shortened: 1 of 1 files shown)


BasilHorowt:
Nice! What's the thinking on the default threads being so high vs chia's defaults?  Wonder if it is more intuitive to have all the pass-through options match up with the chia defaults?

altendky:
I'm not quite sure what to do here.  `:]`  Changelog says 3389 but cli defaults to 4608.  Threads default going down to 4 or 2 (the cli default) makes sense to me.  What do you think @cciollaro?  The suggestion of just tracking the chia defaults sounds good to me.  Less to think about, less to debate.  They're mostly just an example file anyways, not any actual recommendation since that all depends.

cciollaro:
totally agreed. these values are just placeholders and aligning with the command defaults simplifies things. I'll push an update to match the CLI.

altendky (approved):
@cciollaro, thanks for the help getting this up to date.

PR #140, opened 2021-04-20
add support for chia installed via macOS dmg installer

davidtaylorhq:
The official macOS dmg installation makes `chia` available as a standalone binary. That means the `cmdline` does not include `python` in the first position. This commit updates the cmdline parsing logic to allow for `chia` binaries like this.


src/plotman/job.py
```diff
@@ -25,12 +25,13 @@ def job_phases_for_dstdir(d, all_jobs):
...
+    if cmdline and 'python' in cmdline[0]:
+        cmdline = cmdline[1:]
     return (
-        len(cmdline) >= 4
-        and 'python' in cmdline[0]
-        and cmdline[1].endswith('/chia')
-        and 'plots' == cmdline[2]
-        and 'create' == cmdline[3]
+        len(cmdline) >= 3
+        and cmdline[0].endswith("chia")
```
altendky, line 32: Was the `/` removed intentionally?  If so, can you share an example command line from `psutil` where there is no slash?

```suggestion
        and cmdline[0].endswith("/chia")
```

README.md
```diff
@@ -181,11 +181,14 @@ To display the current location of your `plotman.yaml` file and check if it exis
...
+Installation for Linux and macOS:
 
 1. Plotman assumes that a functioning [Chia](https://github.com/Chia-Network/chia-blockchain)
-   installation is present on the system. Activate your `chia` environment by typing
-   `source /path/to/your/chia/install/activate`.
+   installation is present on the system.
+      - Linux: Activate your `chia` environment by typing
+        `source /path/to/your/chia/install/activate`.
+      - macOS: Follow [these instructions](https://github.com/Chia-Network/chia-blockchain/wiki/CLI-Commands-Reference#mac)
+        to add the `chia` binary to the PATH
```
altendky, line 191: Can't chia be used on macOS in the same way it is in Linux?  If I'm not wrong, then I would consider the dmg approach an extra macOS specific option.

```suggestion
      - virtual environment (Linux, macOS): Activate your `chia` environment by typing
        `source /path/to/your/chia/install/activate`.
      - dmg (macOS): Follow [these instructions](https://github.com/Chia-Network/chia-blockchain/wiki/CLI-Commands-Reference#mac)
        to add the `chia` binary to the `PATH`
```

altendky (commented):
Hi @davidtaylorhq, thanks for the help expanding macOS support.  Have a look at my comments and let me know if I'm on to something...  or if I just don't know my macOS details.

src/plotman/job.py
davidtaylorhq: Yeah it was intentional - I should have mentioned that in the commit message. When running the `chia` from the PATH, it looks like this, with no leading slash:
```
% ps -axf | grep chia
  501 25158 25147   0  9:56am ttys003    0:13.87 chia plots create
```

README.md
davidtaylorhq: Yeah you're right - this is much better wording, thanks.

src/plotman/job.py
altendky: Alrighty, sounds good for now.  The pure string methods approach is a bit insufficient.  It should probably be a regex like `"[$/]chia^"` or something.  But, that can be addressed differently.  I want to avoid stuff like scripts named `plot_chia` being identified as chia plotting processes.  Unlikely given all the other checks, but still.  Anyways, as mentioned, improving that is a separate task I'll take on if I really want to.  `:]`

altendky:
Sorry, thought this was a quick catch up merge and then I'd merge it.  But I overlooked that it was targeting `main` instead of `development` so...  dealing with that again.

PR #142, opened 2021-04-21
Update process detection logic to work on OSX

BasilHorowt:
the arguments passed to this function looked like this on OSX:

`['/Library/Frameworks/Python.framework/Versions/3.9/Resources/Python.app/Contents/MacOS/Python', '/Users/chia/chia-blockchain/venv/bin/chia', 'plots', 'create', '-k', '32', '-r', '2', '-u', '128', '-b', '4608', '-t', '/Volumes/plotting2tb', '-d', '/Volumes/16tb']`

The first argument did not contain a lowercase 'python', so we can lowercase the string before comparison to make the check work more generally.

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -27,7 +27,7 @@ def job_phases_for_dstdir(d, all_jobs):
 def is_plotting_cmdline(cmdline):
     return (
         len(cmdline) >= 4
-        and 'python' in cmdline[0]
+        and 'python' in cmdline[0].lower()
         and cmdline[1].endswith('/chia')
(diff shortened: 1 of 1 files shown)


BasilHorowt:
Adjusts the same logic for difference reasons as https://github.com/ericaltendorf/plotman/pull/140/files so one of these will need to be resolved

altendky (approved):
@BasilHorowt, it's good to have a few people helping us improve our macOS support.  Thanks!

altendky:
Just a little paperwork here...  just a sec.

PR #164, opened 2021-04-25
Update default buffer option to 3389 to match Chia default

BasilHorowt:
As per the [previous PR](https://github.com/ericaltendorf/plotman/pull/139), Plotman's default config should track the default Chia options to make things simpler.  As per 1.1.2, Chia actually defaults to 3389 for the buffer: https://github.com/Chia-Network/chia-blockchain/commit/d0649fb57407cfffbaec7baeba86fd76ecb51770


BasilHorowt:
Is it correct that this PR targets development branch?

altendky:
Yes, development is the correct target.  And thanks for this.  `:]`

PR #170, opened 2021-04-26
fix: support -a, --alt_fingerprint parameter

ilazycat:


altendky (approved):
This will be replaced by https://github.com/ericaltendorf/plotman/pull/116 when we get it ready and merged, but in the mean time this looks good.  Thanks!

PR #173, opened 2021-04-27
Fix plotman archive command

cmeyer90:
Small fix for `plotman archive` command. I moved the common code into archive.py and retained the original functionality.


altendky:
@cmeyer90 thanks for this.  Since we don't have tests covering this...  could you explain what testing you have done?  I'm not hoping for much more than something like 'i ran the archive command and it archived a file and it also still archives successfully when running interactive'.

Pending that, I think it looks good.  Long term I'd like to pull the subprocessing out higher, but this seems like an appropriate solution to the immediate issue which is exactly what I wanted.  Thanks again.

cmeyer90:
@altendky - yes of course! I was able to do the following:
- `plotman archive` successfully launches the rsync command and archives a plotfile
- upon quitting `plotman archive` and restarting in `plotman interactive` I can see the PID in the top right corner
- otherwise plotman seems to act normally in both modes.

My testing was brief - I switched I back over to my other branch after quickly verifying these pieces. I agree on the subprocessing bit!

altendky (approved):
Thanks a bunch.  It'll be good to have a not-totally-broke subcommand here again.  `:]`

cmeyer90:
It's especially important in my setup - I have a script that (on reboot) cleans up the plotman tmp folders, moves logs to a backup folder, and starts `plotman archive` and `plotman plot` in tmux.

I am glad we were able to get `plotman archive` working together. Fine piece of software, plotman is. Thank you!

ericaltendorf:
thank you, @cmeyer90 !

PR #174, opened 2021-04-27
Improve compatibility and precision of date information in log filenames

BasilHorowt:
There a few reasons that the current log filenames may not be ideal:

1) They contain `:` which breaks in Windows and is shown as `/` in OSX GUI ([#144](https://github.com/ericaltendorf/plotman/issues/144))
2) They don't contain any more precision than seconds, so with many concurrent plots being created (or error situations), naming collisions can occur
3) They don't contain any timezone information which makes them not fully comprehensive

This potentially closes https://github.com/ericaltendorf/plotman/issues/144

Note: As discussed on chat with @altendky, we settled on `+HH_MM` / `-HH_MM` for the timezone part (`±[hh]:[mm]` with the same `:` -> `_` done for times).  it seems that `strftime` provides the offset as one component via `%z`, without the option to use a `HH:MM` format.  For now I have left it without a separator, but it would be trivial to insert a `_` if it is desired. Thoughts?


src/plotman/manager.py
```diff
@@ -101,7 +102,7 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
                 dstdir = max(dir2ph, key=dir2ph.get)
 
             logfile = os.path.join(
-                dir_cfg.log, datetime.now().strftime('%Y-%m-%d-%H:%M:%S.log')
+                dir_cfg.log, pendulum.now().strftime('%Y-%m-%dT%H_%M_%S.%f%z.log')
```
altendky, line 105: I was thinking we would literally use an ISO format and replace the colons.  Doesn't super matter, but since I checked to see what it would look like...  Thoughts?

```suggestion
                dir_cfg.log, pendulum.now().isoformat(timespec='microseconds').replace(':', '_') + '.log'
```

altendky (commented):
Maybe use `.isoformat()`, but otherwise looks good.  Thanks.

src/plotman/manager.py
BasilHorowt: Ah yeah, way nicer 👍

BasilHorowt: Any documentation on preferred way of handling revisions like this? I can use the github "Commit Suggestion" action here and then squash and reopen? Or force-push? At my job we tend to avoid force-pushing if there's been any line-comments

altendky: I default to not modifying history etc.  I just keep committing and going and so on.  And yeah, I use the suggestion so people can just click the button to commit if they like it.  Shouldn't be any need to squash etc.

altendky (approved):
Cool, cool.  Thanks for the design and coding work.

PR #187, opened 2021-04-29
clean up lines for better format

jjhesk:
added update script for automatic updates.


altendky:
There are a lot of changes going on here.  I expect to switch to black to handle formatting at some point, but not right now.  For the other pieces, if you want to submit either separate PRs or file tickets then we can discuss them.  Sorry.  `:|`

ericaltendorf:
As Kyle said, we'll address code formatting with an automated system at some point, so we'd prefer to not include those changes here.  The rest of the changes should be split into separate PRs to be discussed and approved.  Also, we're developing on branch `development`, so changes should be based off of that.

I've just put together this: https://github.com/ericaltendorf/plotman/wiki/Contributing since we didn't really have a guide before :)

Thanks for understanding.

altendky:
It seems that this has turned into a fork.  It would be helpful if you could make that very obvious everywhere it is published to help avoid confusion.  Also update all links etc in the fork.

Since this isn't really a PR at this point I'm going to close it to avoid notifications.  (maybe?)  If you would like your work included here feel free to submit focused PRs.

PR #190, opened 2021-04-29
Edit default archival config to add helpful documentation

artorantala:
I had a confusing time setting up the archiving. I added some key comments that would have been very useful to me in figuring out how the archival works.


ericaltendorf (approved):

altendky (approved):
@artorantala, thanks for filling this out.  There is certainly a lot of confusion around this setup and it is very understandable that people have trouble.  Hopefully this will save a few people some time.

PR #193, opened 2021-04-30
fix: archiving status could be None

altendky:
```python-traceback
  File "/home/altendky/.local/bin/plotman", line 33, in <module>
  File "/farm/plotman/src/plotman/plotman.py", line 171, in main
    interactive.run_interactive()
  File "/farm/plotman/src/plotman/interactive.py", line 340, in run_interactive
    curses.wrapper(curses_main)
  File "/home/altendky/.pyenv/versions/3.9.1/lib/python3.9/curses/__init__.py", line 94, in wrapper
    return func(stdscr, *args, **kwds)
  File "/farm/plotman/src/plotman/interactive.py", line 234, in curses_main
    archiving_status_msg(archiving_configured,
  File "/farm/plotman/src/plotman/interactive.py", line 59, in archiving_status_msg
    return '(active) ' + status
TypeError: can only concatenate str (not "NoneType") to str
```

FYI, @cmeyer90.  Looks like we both missed something over in #173.  `:]`

Draft for:
- [x] At least some minimal testing here where the exception just happened.

### DIFF
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -15,12 +15,12 @@ from plotman import manager, plot_util
 
 # TODO : write-protect and delete-protect archived plots
 
-def spawn_archive_process(dir_cfg, all_jobs, previous_archiving_status):
+def spawn_archive_process(dir_cfg, all_jobs):
     '''Spawns a new archive process using the command created 
     in the archive() function. Returns archiving status and a log message to print.'''
 
     log_message = None
-    archiving_status = previous_archiving_status
+    archiving_status = None
     
     # Look for running archive jobs.  Be robust to finding more than one
     # even though the scheduler should only run one at a time.
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -15,12 +15,12 @@ from plotman import manager, plot_util
 
 # TODO : write-protect and delete-protect archived plots
 
-def spawn_archive_process(dir_cfg, all_jobs):
+def spawn_archive_process(dir_cfg, all_jobs, previous_archiving_status):
     '''Spawns a new archive process using the command created 
     in the archive() function. Returns archiving status and a log message to print.'''
 
     log_message = None
-    archiving_status = None
+    archiving_status = previous_archiving_status
     
     # Look for running archive jobs.  Be robust to finding more than one
     # even though the scheduler should only run one at a time.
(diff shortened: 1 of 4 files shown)


cmeyer90:
😬 will work on this right now, looks like it could be an easy fix. will write some tests as well. @altendky

altendky:
@cmeyer90 I've got 'a' fix here.  I'm wondering if we should move away from the memory and just duplicate/reuse the `archiving_status = 'pid: ' + ', '.join(map(str, arch_jobs))` line after starting the process though.  As is it waits for the next call to report the pid.

altendky:
That looks a bit less intrusive to me and reduces latency by a cycle.  I've got a plot about to archive now so it'll get exercised in a few minutes.

altendky:
hmm, it got the `pid:` bit but without any PIDs then filled one in the next time.  Maybe you do have to wait a bit for the process to get listed even after the `Popen` call.  Or...  since we use `shell=True` (to be stopped at some point) the `bash` process may have been starting up and in the process list but it had not yet gotten around to creating the sub-sub-process for `rsync`.  Hummm...

cmeyer90:
I was quickly phone posting and only looked at the error! I see this is a fix, yep.

I like the idea of showing the previous archive status, but the second commit (c89356d) is much much cleaner. I was thinking this is a possibility as well, in interactive.py on line 57 (I can't figure out how to make it bold in the code block, but I have added **and status is not None**):

```
def archiving_status_msg(configured, active, status):
    if configured and status is not None:
        if active:
            return '(active) ' + status
        else:
            return '(inactive) ' + status
    else:
        return '(not configured)'
```

That way, if there is no status, we go straight to 'not configured' and hopefully avoid the NoneType error. I could be missing some important logic but I think this may work.

altendky:
The `None` was happening when the archiving process was launched I think so `not configured` wouldn't be appropriate in that case.  Faking a plot to see how this goes with `<pending>`.  Probably some better way to make an empty 100GB file but `dd if=/dev/zero` will cut it for now I guess.

altendky:
Seemed to work ok.  Let me know how you feel about it.  Not the most beautiful thing ever, but perhaps isolated enough that it is ok.

cmeyer90:
I do like the pending status @altendky. I spent a little more time thinking about other ways to handle, but this works well. <pending> is an accurate way of describing what's going on as we wait for the process to appear. 

I would imagine at some point this section gets refactored and if that's the case it is good to stick with our original mindset of a quick fix. Looks clean enough to manage in the short term!

altendky:
Thanks for the discussion and review.  I'll go ahead and merge this in.  And yes, I haven't thought about exactly how it should change but this does feel like it could use a refactor at some point.  For now I'll keep squashing bugs and trying to get a few more infrastructure pieces in.

PR #203, opened 2021-05-01
Add `tmpdir_stagger_phase_limit` to overrides

jcronyn:
This PR adds `tmpdir_stagger_phase_limit` to `TmpOverrides` and updates `phases_permit_new_job` as needed. This allows setting different phase limits for different size/speed drives.

Have been testing this and it works as expected.

### DIFF
--- a/src/plotman/configuration.py
+++ b/src/plotman/configuration.py
@@ -50,6 +50,7 @@ class Archive:
 @dataclass
 class TmpOverrides:
     tmpdir_max_jobs: Optional[int] = None
+    tmpdir_stagger_phase_limit: Optional[int] = None
 
 @dataclass
 class Directories:
--- a/src/plotman/manager.py
+++ b/src/plotman/manager.py
@@ -50,18 +50,23 @@ def phases_permit_new_job(phases, d, sched_cfg, dir_cfg):
     if len(phases) == 0:
         return True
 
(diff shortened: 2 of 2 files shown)


erdincsk:
Hello, sorry, i'm just trying to understand. at the moment, tmpdir_stagger_phase_limit limits running plots  in all temp dirs not individually.

is that implementation for to use tmpdir_stagger_phase_limit parameter for each tmp directory.?

thanks.

jcronyn:
@erdincsk - Yes. The concept is that if you have different drives that are capable of different number of parallel plots, you may want to have a different number in tmpdir_stagger_phase_limit (ideally, all fields would be overridable including the major and minor phases, but this is a start)

XiaoLuxtl:
commit please

jcronyn:
@ericaltendorf - I noticed that I neglected to update the documentation (config template) to reflect this update. Do you want me to update that or are you planning to?

altendky:
It looks like https://github.com/ericaltendorf/plotman/pull/758 and https://github.com/ericaltendorf/plotman/pull/427 both provide this as well as the other pieces of the phase override so I expect to go with one of them.  I apologize that I have failed to keep up with the community contributions and I do appreciate your effort to make the changes and submit them here.

PR #204, opened 2021-05-01
include basic usage instructions

drewmullen:
include note on issue when installing plotman on rhel via chia's guide

include basic usage steps


README.md
```diff
@@ -200,6 +200,20 @@ Installation for Linux:
...
 
+*Note:* If you see `ModuleNotFoundError: No module named 'readline'` when using `plotman` on [RHEL based linux](https://github.com/ericaltendorf/plotman/issues/195) after installing using [chia's guide](https://github.com/Chia-Network/chia-blockchain/wiki/INSTALL#centos--red-hat--fedora), install `readline-devel` then reinstall chia starting at compiling python in a new build environment; or consider using a project like `pyenv`.
+
+## Basic Usage:
+
+1. Install
+
+2. Generate initial config
+
+3. Configure (default location ~/.config/plotman/plotman.yaml). Options explained in the default config file (step 2)
```
altendky, line 211: This location is OS specific.  `plotman config path` will tell you the actual path that will be looked for.

README.md
```diff
@@ -200,6 +200,20 @@ Installation for Linux:
...
+
+4. Create log directory specified in `directories: { log: "" }`
+
+5. `plotman plot`
```
altendky, line 215: Do we want to encourage `plot` or `interactive`?  I'm not really sure...

altendky (changes requested):
Thanks for the follow up here.  Sorry for the wait on the review.

ericaltendorf:
regarding `plot` vs `interactive`, I think we can suggest both, each with a
one-liner saying what it does.  i wonder if some people are using `plot`
just to avoid the archiving that `interactive` does by default?

altendky (approved):
Ugh, I need to start processing from the end of the PR least, not the beginning...  to avoid leaving PRs like this languishing.  My apologies.

I'm going to call this good.  If you want to add more, feel free to follow up with another PR.  Hopefully I find dev time soon so we can get away from tmux being useful but for now people seem to like it (along with screen).

PR #210, opened 2021-05-02
Allow empty buffer and add a new archive method

wisherhxl:
Major:
1. set dst directory to optional, if directory.dst is None, no buffer will be used, chia plot command will set dst directory to tmp directory. if multiple fast M.2(>3000M/s) ssd is used and not raid, archive directly from the tmp directory can save disk and time.
2. add archive method by simply move plot to another folder. rsync is greate on linux, but if target famer is windows, rsync can be a pain, if smb or nfs is attached, just move to the mounted directory is ok. also, move then rename can avoid farmer error by select plot file too early.

Minor
1. add sleep time setting for archive thread, optional and default to 60.
2. add status and reason for every archive run.

Cons:
1. To add archive method "move", the Archive class is altered, so old config file have to be changed.


altendky:
From the description, this sounds like four independent changes.  Does that sound right?  If so, it will be a lot easier to get them reviewed and merged if they are done separately.  For archiving options, we are already talking about how to handle that in #88 so having a third option should easily be added in terms of the config once we figure that out.

wisherhxl:
> From the description, this sounds like four independent changes. Does that sound right? If so, it will be a lot easier to get them reviewed and merged if they are done separately. For archiving options, we are already talking about how to handle that in #88 so having a third option should easily be added in terms of the config once we figure that out.

PR #223, opened 2021-05-03
feat: collapse split tmpdir tables into one table

ericaltendorf:


altendky (approved):
reads fine, i trust it ran

Issue #228, opened 2021-05-04
Memory usage display appears to be higher than actual


BasilHorowt:
There are potentially two things here, let me know if they should be broken out separately.

1) psutil memory_info returns values in bytes, and I believe ram (unlike drive space) is still 1024-based, but plotman's [human_format](https://github.com/ericaltendorf/plotman/blob/c0fe10cf09e6b1d9ae04101e7babc1229d0c740f/src/plotman/plot_util.py#L15) appears to be hardcoded to be dealing with drive space and doing a division by 1000. In my case I see a value of 422G whereas my os reports 393G, These would line up if `human_format` was using 1024 `393 / 1000 / 1000 / 1000 * 1024 * 1024 * 1024 = 422`

2) Right now plotman is using [`memory_info.vms`](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info).  On OSX (at least) this value is hugely inflated.  For one Python process I see these value in my "Activity Monitor":

Real Memory Size: 2.42GB
Virtual Memory Size: 393.39GB
Shared Memory Size: 12.8MB
Private Memory Size: 2.41GB

I have absolutely no idea why VM size is reported this high, overall I have 11GB out of 16GB used, and a stated swap of 2GB.

I noticed that there is a `[memory_full_info](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info)` that can report uss:

>Note uss is probably the most representative metric for determining how much memory is actually being used by a process. It represents the amount of memory that would be freed if the process was terminated right now.

Would `memory_full_info().uss` be a more useful metric to use?

BasilHorowt:
I think #2 here is likely mainly quirk of OSX however I did find this

>hmm i just realized, I told plotman in it's cfg to use 3600 ram and 2 cores, but looking at plotman interactive i see each plot us using around 4.2gb of ram? is that normal?

- [Keybase](keybase://chat/chia_network.public#plotman/12535)

The 1000 vs 1024 issue alone doesn't explain 4.2gb being reported for 3600, so perhaps `memory_info().vms` is off on linux too?

thauburger:
Just as another concrete example here, I'm seeing the same large MEM gap on macOS (10.15)

For my plotman configuration, I'm assigning the minimum required RAM to maximize parallel plotting on slower HDD disks.

`-b 2500`

Here's a photo of my chia plotting processes in `top`:

[...]

PR #229, opened 2021-05-04
Display memory usage as factor of 1024 rather than 1000

BasilHorowt:
I think this is expectation for memory usage and makes the numbers line up with what the OS reports


BasilHorowt:
Potentially closes part of https://github.com/ericaltendorf/plotman/issues/228

BasilHorowt:
Apologies, I am not able to run `pytest` ("fixture 'mocker' not found"), might not have nailed the tests

altendky:
`myenv/bin/pip install --editable .[dev]` should get the deps.

src/plotman/plot_util.py
```diff
@@ -12,11 +12,14 @@ def df_b(d):
 def get_k32_plotsize():
     return 108 * GB
 
-def human_format(num, precision):
+def human_format(num, precision, powerOfTwo=False):
     magnitude = 0
     while abs(num) >= 1000:
```
altendky, line 17: This needs to vary as well.  Maybe start with `divisor = 1024 if power_of_two else 1000`.

altendky (changes requested):
@ericaltendorf opinions?  I wonder about adding the full `GB` and `GiB`.  `df` uses `T` and such and I don't know which it means or if it is 'universal' or just presumed due to being a disk utility...  `free` does use `Gi`, so maybe that's the 'short' way.  `G` and `Gi`.  Could just add an `i` if `power_of_two`.

altendky:
Oops, said not ready for review...  Well, it seemed close enough to be worth reviewing I guess.  `:]`

BasilHorowt:
> Oops, said not ready for review... Well, it seemed close enough to be worth reviewing I guess. `:]`

Oh sorry! I am ready for review... just not confidant in any way in my python so I guess I was using "Draft" as ... "This might be a decent way to do it". Thank you for the review!

src/plotman/plot_util.py
BasilHorowt: ah yes, will update

BasilHorowt: ah yes, will update

BasilHorowt:
>myenv/bin/pip install --editable .[dev] should get the deps.

@altendky 

Thank you, you mentioned that on keybase too but i never got back to you, basically i get a weird error:

> pip install --editable .[dev]
zsh: no matches found: .[dev]

However I installed the [dev deps ](https://github.com/ericaltendorf/plotman/blob/5788e7fa2df2186ccb94db2d388766f9d52116a4/setup.cfg#L58) manually and now `pytest` works

src/plotman/plot_util.py
```diff
@@ -12,11 +12,13 @@ def df_b(d):
...
+    divisor = 1024 if powerOfTwo else 1000
+    
     magnitude = 0
-    while abs(num) >= 1000:
+    while abs(num) >= divisor:
         magnitude += 1
-        num /= 1000.0
+        num /= divisor        
     return (('%.' + str(precision) + 'f%s') %
             (num, ['', 'K', 'M', 'G', 'T', 'P'][magnitude]))
```
altendky, line 23: Sorry this got left hanging...  Does this seem ok to you?  Does seem like it should be indicated.

```suggestion
    result = (('%.' + str(precision) + 'f%s') %
            (num, ['', 'K', 'M', 'G', 'T', 'P'][magnitude]))

    if powerOfTwo and magnitude > 0:
	    result += 'i'
    
    return result
```

src/plotman/_tests/plot_util_test.py
```diff
@@ -11,6 +11,8 @@ def test_human_format():
     assert (plot_util.human_format(3542000, 2) == '3.54M')
     assert (plot_util.human_format(354, 0) == '354')
     assert (plot_util.human_format(354, 2) == '354.00')
+    assert (plot_util.human_format(422399296143, 2) == '422.40G')
+    assert (plot_util.human_format(422399296143, 2, True) == '393.39G')
```
altendky, line 15:
```suggestion
    assert (plot_util.human_format(422399296143, 2, True) == '393.39Gi')
```

src/plotman/_tests/plot_util_test.py
```diff
@@ -11,6 +11,8 @@ def test_human_format():
     assert (plot_util.human_format(3542000, 2) == '3.54M')
     assert (plot_util.human_format(354, 0) == '354')
```
altendky, line 12:
```suggestion
    assert (plot_util.human_format(354, 0) == '354')
    assert (plot_util.human_format(354, 0, True) == '354')
```

altendky (changes requested):

ericaltendorf (approved):

altendky:
I got the go ahead from Eric, so once we have the `i` discussion done this is good to go.  Thanks as always.

PR #233, opened 2021-05-04
add sleep time for archive

wisherhxl:
add sleep time setting for archive thread, optional and default to 60 and for every archive run, show status.

### DIFF
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -130,7 +130,7 @@ def archive(dir_cfg, all_jobs):
     if not archdir:
         return(False, 'No archive directories found with enough free space')
     
-    msg = 'Found %s with ~%d GB free' % (archdir, freespace / plot_util.GB)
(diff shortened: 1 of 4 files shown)


altendky:
Presently, in interactive mode archiving is periodic at `polling_time_s`.  This PR would make archiving occur at different periods depending on running `plotman archive` vs. `plotman interactive`.  It seems like this would be confusing.  What do you think of using the existing `polling_time_s` in all cases?  Otherwise I think I would want to change interactive to follow any new archiving check period.  I can imagine how polling the two differently might make sense but I do wonder how much value it brings for the 'extra complexity'...  I mean, it's not that complex, but does it bring significant value?

What do you think of all that?

And, thanks for your work here.  We appreciate all the contributions people provide.  `:]`

wisherhxl:
the sleep time for archive was set to 60 in the source code, that's why I create this PR.
I will close it and checkout the dev branch, thank you!

PR #234, opened 2021-05-04
allow empty dst directory, if so, the tmp directory is used as buffer.

wisherhxl:
set dst directory to optional, if directory.dst is None, no buffer will be used, chia plot command will set dst directory to tmp directory.
If multiple fast M.2(>3000M/s) ssd is used and not raid, archive directly from the tmp directory can save disk and time.


src/plotman/configuration.py
```diff
@@ -35,6 +35,13 @@ def get_validated_configs():
     except marshmallow.exceptions.ValidationError as e:
         raise ConfigurationException(f"Config file at: '{config_file_path}' is malformed") from e
 
+def get_dst_directories(dir_cfg):
+    """Returns either (True, <Directories.dst>) or (False, <Directories.tmp>). If Directories.dst is None,
+    Use Directories.tmp as dst directory."""
+    if dir_cfg.dst is None:
+        (False, dir_cfg.tmp)
+    else:
+        (True, dir_cfg.dst)
```
altendky, line 44: What do you think of making this a method or two on `Directories`?  Maybe `dir_cfg.dst_is_tmp()` and `dir_cfg.effective_dst_dirs()` or such?

Also, as is, this is missing returns for the specified tuples.

altendky (changes requested):
Thanks for the work here, it looks like a good addition.  It half applies to me right now (the dst-on-tmp part) and maybe fully next week depending if I RAID in my new system or not (the multiple tmp drives part).  If you could test it out for a bit to make sure it works, that would be great.

wisherhxl:
> Thanks for the work here, it looks like a good addition. It half applies to me right now (the dst-on-tmp part) and maybe fully next week depending if I RAID in my new system or not (the multiple tmp drives part). If you could test it out for a bit to make sure it works, that would be great.

I'm using multiple M.2 SSD and use them seperately, so you will be fine. using tmp dir as dst can save the copy time, but not recommand for HDDs.

altendky:
I don't know if we just lost each other or what, but I hope you don't mind that I'm going to catch this up and finish it out.  Thanks so much for the contribution.  I look forward to being able to use this whenever I finally get my new setup...  `:]`

altendky (approved):
I've been running this for a day or two without issue.  Excited to get another temp setup so I need this.  `:]`  Thanks again for the work here @wisherhxl.

7AC:
Is there an ETA for a release with this?

altendky:
Maybe a week?  Other than trying to kind of not really keep up with reported issues etc, I have a custom archiving change that I'd like to get out.  You can use the regular installation instructions and replace `@main` with `@development` to use it now.

7AC:
Thanks appreciate it.

Issue #236, opened 2021-05-04
plotman kill not working


kxknet:
I use newest version chia and latest developer version plotman
See this
plotman kill 2c3e4235
Namespace(cmd='kill', idprefix=['2c3e4235'])
Pausing PID 4195, plot id 2c3e4235e0cf36fcfeb167b7da60f4310a6a47a6c4545011a8573d2120041d83
Traceback (most recent call last):
  File "/chia-blockchain/venv/bin/plotman", line 8, in <module>
    sys.exit(main())
  File "/chia-blockchain/venv/lib/python3.8/site-packages/plotman/plotman.py", line 230, in main
    temp_files = job.get_temp_files()
  File "/chia-blockchain/venv/lib/python3.8/site-packages/plotman/job.py", line 373, in get_temp_files
    if self.tmpdir in f.path or self.tmp2dir in f.path or self.dstdir in f.path:
TypeError: 'in <string>' requires string as left operand, not NoneType

JordyBaylac:
@kxknet it was happening to me as well, I  have submitted a PR

braveckin:
It "works" for me in the way that it suspends the job instead of killing it and removing the temp files as it states in the message... I was even able to resume the job...

PR #238, opened 2021-05-04
fix plotman kill when temp2dir is not set

JordyBaylac:
This PR is a fix to plotman kill command when temp2 directory is not specified in config. It is failing as specified in Issue #236.

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -1,6 +1,7 @@
 # TODO do we use all these?
 import argparse
 import contextlib
+import functools
 import logging
 import os
 import random
@@ -11,6 +12,7 @@ from datetime import datetime
 from enum import Enum, auto
 from subprocess import call
 
+import attr
 import click
 import pendulum
 import psutil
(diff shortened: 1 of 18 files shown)


src/plotman/job.py
```diff
@@ -370,7 +370,9 @@ def get_temp_files(self):
         # Prevent duplicate file paths by using set.
         temp_files = set([])
         for f in self.proc.open_files():
-            if self.tmpdir in f.path or self.tmp2dir in f.path or self.dstdir in f.path:
+            if self.tmpdir in f.path or self.dstdir in f.path:
+                temp_files.add(f.path)
+            if self.tmp2dir is not None and self.tmp2dir in f.path:
```
altendky, line 375: What do you think of this?

```suggestion
            if any(
                dir in f.path
                for dir in [self.tmpdir, self.dstdir, self.tmp2dir]
                if dir is not None
            ):
```

altendky (commented):
Thanks for the work fixing this 'corner case'.  We've got plenty of them and some of them aren't all that 'cornery'...  It's good to get some help with them.

altendky:
I'm doing too many things...  It appears that the fix I suggested already made it into `development` at some point.  Thank you for your contribution and Hopefully we don't end up crossing efforts next time.

PR #257, opened 2021-05-07
run in vscode

Ed-Yang:
Make a little change to support VSCode debugging.

1. While debugging in the vscode IDE on macOS, the process name of Python it "Python" not "python", so add additional check.

2. In addition, if the filename is unicode, the decode will cause exception, so change it from 'ascii' to 'utf-8'.

It can test with the following configuration in launch.json (in chia-blockchain workspace):

{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Python: plotman",
            "type": "python",
            "request": "launch",
            "module": "plotman",
            "justMyCode": false,
            "args":["interactive"]
        }
    ]
}

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 __pycache__
 venv
+.DS_Store
+.vscode
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -58,7 +58,7 @@ def get_archdir_freebytes(arch_cfg):
                 # not actually mounted
                 continue
             freebytes = int(fields[3][:-1]) * 1024  # Strip the final 'K'
-            archdir = (fields[5]).decode('ascii')
+            archdir = (fields[5]).decode('utf-8')
             archdir_freebytes[archdir] = freebytes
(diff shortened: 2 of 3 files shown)


altendky (changes requested):
Thanks for this.  I think the `utf-8` bit is good.  The Python capitalization has already been addressed on `development`.  This PR should also be targeted at the `development` branch.

https://github.com/ericaltendorf/plotman/blob/d13628aaaad580ce70133648d4e882e79f2ddb82/src/plotman/job.py#L30
https://github.com/ericaltendorf/plotman/blob/d13628aaaad580ce70133648d4e882e79f2ddb82/src/plotman/job.py#L46

Ed-Yang (commented):
Thanks!

Ed-Yang:
some issues already be addressed, I close this and re-issue one only cover unicode filename instead.

PR #262, opened 2021-05-07
run in vscode

Ed-Yang:
As requested, issue a PR on development branch.


Ed-Yang:
some issues already be addressed, I close this and re-issue one only cover unicode filename instead.

PR #271, opened 2021-05-08
attrs

altendky:


altendky:
Use attrs because...  that's what I do.  Frozen instances until we find some need to mutate and think about it carefully.

PR #277, opened 2021-05-08
support K25

Ed-Yang:
Support k25 plot size to shorten development cycle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 __pycache__
 venv
+.DS_Store
+.vscode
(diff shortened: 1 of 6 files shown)


Ed-Yang:
test failed....

Ed-Yang:
I am not familiar with Git, this PR also include changes made in PR #264.  If that make any inconvenient, feel free to modify or drop both PRs.  Thanks.

altendky:
Multiple developers and users want this feature, myself included, but we'll need some more work on it.  My general plan is to have it be available only as a command line option, not in the configuration file, and to provide a strong warning when executing either `plotman interactive` or `plotman plot`.  I very much want to avoid people accidentally starting plotting k25s and then coming back a couple days later with 10TiB of k25s they have to delete.  `:]`

Ed-Yang:
close

PR #299, opened 2021-05-09
monitor

altendky:


src/plotman/monitor.py
```diff
@@ -0,0 +1,243 @@
...
+                    jobs=jobs,
+                    dst_prefix=dst_prefix,
+                    tmp_prefix=tmp_prefix,
+                )
+
+                jobs_table = build_jobs_table(jobs_data=jobs_data)
+                with rich_console.capture() as capture:
+                    rich_console.print(jobs_table)
+
+                plots_buffer.text = capture.get()
```
altendky, line 165: I get output like this instead of black on white text.  Any ideas how to get the fancy output from `rich` into `prompt_toolkit` properly?

``^[[7m ^[[0m^[[7m#^[[0m^[[7m ^[[0m^[[7m ^[[0m^[[7mplot_id ^[[0m^[[7m ^[[0m^[[7
```

PR #305, opened 2021-05-10
New configuration option for interactive

rafaelsteil:
## What does this change accomplish?
Add a new configuration option to `interactive` and a command line argument to enable or disable automatic plotting when opening the UI. 

Prior to that, anytime `plotman interactive` ran it would automatically fire any pending jobs, even before the user had the chance to press `p` to change its status. 

This PR introduces two new options:
1. A new section in `plotman.yaml` 
2. A command line argument

The section is `tools/interactive/autostart_plotting`, which defaults to True in order to keep backwards compatibility. The command line argument is `--autostart-plotting` and `--no-autostart-plotting`, which take precedence over `plotman.yaml``

## How to test it
1. Set `tools/interactive/autostart_plotting` to `False` and later to `True` and check the `interactive ` reflects the plotting status
2. Pass the command line arguments and check if they override what was set on the configuration file
3. Remove `tools` from the configuration file and run `plotman interactive`. It should show plotting as active (the original behaviour)


README.md
```diff
@@ -147,6 +147,25 @@ archive jobs initiated.  This is the one part of the interactive tool which is
 stateful.  There is no permanent record of these executed command lines, so if
 you start a new interactive plotman session, this log is empty.
 
+## `plotman` commands
```
altendky, line 150: Don't move it now, but this would generally be good as a separate PR since it is mostly unrelated.  I can see a relationship to the passive mode being added to interactive, though it isn't pointed out here.  So maybe that's the real comment, to note that along with running plot as a daemon you can then set interactive to not autostart plotting.

src/plotman/plotman.py
```diff
@@ -30,7 +30,16 @@ def parse_args(self):
...
+        p_interactive = sp.add_parser('interactive', help='run interactive control/monitoring mode')
+        p_interactive.add_argument('--autostart-plotting',
+            action='store_true',
+            default=None,
+            help='Automatically start plotting when the tool opens')
+
+        p_interactive.add_argument('--no-autostart-plotting',
+            action='store_true',
+            default=None,
+            help='Do not start plotting when the tool opens')
```
altendky, line 42: Take a look at `BooleanOptionalAction` in https://docs.python.org/3/library/argparse.html#action.  I think this can be a single added argument.

src/plotman/configuration.py
```diff
@@ -94,9 +94,18 @@ class Plotting:
 class UserInterface:
     use_stty_size: bool = True
 
+@attr.frozen
+class Interactive:
+    autostart_plotting: bool
+
+@attr.frozen
+class Tools:
+    interactive: Optional[Interactive] = None
```
altendky, line 103: Could we just give a default value here?  Or do we need to explicitly know when this isn't specified.

```suggestion
@attr.frozen
class Interactive:
    autostart_plotting: bool = True

@attr.frozen
class Tools:
    interactive: Interactive = attr.ib(factory=Interactive)
```

altendky (changes requested):
Hey @rafaelsteil, thanks for jumping in and getting this going.  I have definitely wanted this myself in various senses.  There should be a similar option for archiving.  It could fit in this same PR if you want.  Or we could handle it separately afterwards.

src/plotman/resources/plotman.yaml
```diff
@@ -9,6 +9,13 @@ user_interface:
         # you resize the terminal window, you can try setting this to True. 
         use_stty_size: True
 
+# Optional custom settings for the tools (status, interactive etc)
+tools:
```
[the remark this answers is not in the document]
rafaelsteil: Yeah, makes sense. I'll update it

src/plotman/plotman.py
altendky: Slight update to account for the not-passed case so we can use the value in the config file.

https://replit.com/@altendky/ImmaterialTintedInterfacestandard-2

```python
import argparse


parser = argparse.ArgumentParser()
parser.add_argument('--x', action='store_true', default=None, dest='x')
parser.add_argument('--no-x', action='store_false', default=None, dest='x')

print(parser.parse_args([]).x, None)
print(parser.parse_args(['--x']).x, True)
print(parser.parse_args(['--no-x']).x, False)
print(parser.parse_args(['--x', '--no-x']).x, False)
print(parser.parse_args(['--no-x', '--x']).x, True)
```

```
None None
True True
False False
False False
True True
```

PR #309, opened 2021-05-10
Plotman doesn't work if installed with experimental gui installer

bradsquicciarini:
## Plotman can't find jobs if it was installed with experimental gui installer

This has lots of unfortunate consequences like infinitely launching jobs :upside_down_face: 

+ Adds a `gui_install` flag to config file
+ Modifies `jobs.py` to search for jobs differently if the flag is set

## Testing (?)

Not sure how to test. Let me know the preferred way and I can run them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__
 venv
+src/plotman.egg-info/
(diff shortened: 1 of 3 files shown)


altendky:
Also, PRs should target the development branch.

Issue #321, opened 2021-05-10
[req] Add option to treat archive destination as directory instead of drives


Fmstrat:
I know work is ongoing with archiving, but until then the one holdup for me is the fact that the archiver expects there to be drives inside the shared directory. In my case, I'm archiving to a sub-directory in a ZRAID, so the `df` always returns nothing.

I'm in testing on a PR to add a fix for this now.

PR #322, opened 2021-05-10
req: allow archive to send to directory vs drives

Fmstrat:
Handles https://github.com/ericaltendorf/plotman/issues/321

The archiver currently requires separate drives to be mounted for archiving. For instance:

- /mnt (a folder on the root drive)
  - /mnt/drive1 (a physical drive/mount in `df`)

However, if you have a RAID or ZRAID where you are storing elements, there is no "mounted drive" to use. For instance:

- /storage (a RAID array with lots of items on it
  - /storage/random (some random files)
  - /storage/plots (where you want your plots to go)
  - /storage/morerandom (some random files)

This PR adds a new configuration option of `rsyncd_type` which defaults to `drives` if it does not exist in the configuration (the original method).

If it is set to `directory` in the configuration YAML, then:
- `df` checks free space on the destination directory without requiring it to be a separate mounted disk
- The same path replacements happen, which means the share name (ex: `chia`) would be the rsync destination, which in turn is the single directory destination

I've run this across 3 plots in testing, and all functions smoothly.

### DIFF
--- a/src/plotman/_tests/archive_test.py
+++ b/src/plotman/_tests/archive_test.py
@@ -8,6 +8,7 @@ def test_compute_priority():
 def test_rsync_dest():
     arch_dir = '/plotdir/012'
     arch_cfg = configuration.Archive(
+        rsyncd_type='drives'
         rsyncd_module='plots_mod',
         rsyncd_path='/plotdir',
         rsyncd_host='thehostname',
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -49,6 +49,20 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = {}
+    df_cmd = ("ssh %s@%s df -aBK '%s' 2>/dev/null | awk 'NR==2 {print $4}'" %
+        (arch_cfg.rsyncd_user, arch_cfg.rsyncd_host, arch_cfg.rsyncd_path) )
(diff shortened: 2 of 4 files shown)


src/plotman/_tests/archive_test.py
```diff
@@ -8,6 +8,7 @@ def test_compute_priority():
 def test_rsync_dest():
     arch_dir = '/plotdir/012'
     arch_cfg = configuration.Archive(
+        rsyncd_type='drives'
```
crisperhills, line 11: Might appease tests:
```suggestion
        rsyncd_type='drives',
```

altendky:
Archiving is being significantly reworked in terms of the identification of target directories and the transfer commands.  This will allow users to enter their own scripts for those activities as well as set us up to generically provide multiple presets.  There is already a local and remote rsync option in `target_definitions.yaml` in https://github.com/ericaltendorf/plotman/pull/627, though both use the existing 'find mounted drives in the specified directory' logic.  The interest here should be able to be implemented as another preset.  After the custom archival PR is merged, let's add a preset there.

altendky:
I believe you could implement this now by way of a custom archive target definition (released in v0.4).  The transfer script wouldn't need to change.  The disk space script could loop over the configured directories and check and report the space for each.

I do appreciate the effort to contribute this to the project and apologize for my very slow handling of yours and other contributions.

altendky:
I think that adding this now would take a sufficiently different form that it probably ought to be submitted as a new PR.

PR #330, opened 2021-05-11
Avoid double counting jobs if chia was installed with binaries

bradsquicciarini:
If chia was installed with binaries, plotman will double count jobs. As you can see from the structure from below, there are two processes running for one job. (If someone could explain this, that'd be great)

image

+ Adds a check in `job.py` to only count the child process.

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -149,6 +149,8 @@ class Job:
                             parsed_command = parse_chia_plots_create_command_line(
                                 command_line=proc.cmdline(),
                             )
+                            if proc.parent().name() != 'chia':  # grab thread not parent
+                                continue                            
                             if parsed_command.error is not None:
                                 continue
                             job = Job(
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ __pycache__
 venv
 .DS_Store
 .vscode
+src/plotman.egg-info


rafaelsteil:
> _As you can see from the structure from below, there are two processes running for one job. (If someone could explain this, that'd be great)_

The green ones are the threads, not processes. Phase 1 runs with the number of threads you set (`r 2`), and from phases 2-4 it uses only one thread, that's why some rows show just one `chia` command

altendky:
Side note, the shared screenshot shows two `plotman interactive` commands running which can create multiple plots when otherwise there shouldn't be and for awhile that could result in two plot processes logging to a single file causing the two separate plotting activities to show up with the same plot id in plotman.

src/plotman/job.py
```diff
@@ -149,6 +149,8 @@ def get_running_jobs(logroot, cached_jobs=()):
                             parsed_command = parse_chia_plots_create_command_line(
                                 command_line=proc.cmdline(),
                             )
+                            if proc.parent().name() != 'chia':  # grab thread not parent
+                                continue
```
rafaelsteil, line 153: This will cause other problems, since jobs created by `plotman plot` will return `plotman` as parent's name

altendky:
Fixed in https://github.com/ericaltendorf/plotman/pull/418.  Thanks for the help and submission.

PR #338, opened 2021-05-11
Fix Phase visualizer

lopesmcc:
Plotman interactive phase visualizer broke - just another tuple that needed fixing.


altendky:
Thanks for cleaning up after me here.  Given this is a regression I would like to get it fixed asap and probably into the pending release.  I'll take a look at the broken tests and try to figure this out.

PR #340, opened 2021-05-11
removes unessisary arguments in psutil.pocess_iter in job.Job.get_running_jobs()

hackerzhuli:
The attributes retrived with argumments (that is, 'pid' and 'cmdline') is not used, and getting 'cmdline' attribute with psutil from some processes on windows is unstable and throws exceptions.

It is impossible to catch exceptions from this line without abandoning the for lop. If we try to catch exception from this line, we will not get jobs from processes that comes after the exception, this will cause big problems, that is, running jobs that are not returned from get_running_jobs()!

It is better to get 'cmdline' and other attributes of processes inside the for loop, where exceptions can be handled without abandoning the for loop, that is, we can move on to processes that comes after the process that throws the exception.

Here is info about the exception that can be caused by this line:
OS: Windows 7
python version: 3.8.9
psutil: 5.8.0
This exception is caused by accessing 'cmdline' attribute on process 'audiodg.exe'.
Note that, sometimes 'audiodg.exe' is not present on the system, try to ajust audio volumn in the right of task bar, this will cause 'audiodg.exe' to run.

Here is the traceback:

```python-traceback
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python38\Scripts\plotman.exe\__main__.py", line 7, in <module>
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\plotman\plotman.py", line 162, in main
    jobs = Job.get_running_jobs(cfg.directories.log)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\plotman\job.py", line 140, in get_running_jobs
    for proc in psutil.process_iter(['pid', 'cmdline']):
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\__init__.py", line 1439, in process_iter
    yield add(pid)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\__init__.py", line 1416, in add
    proc.info = proc.as_dict(attrs=attrs, ad_value=ad_value)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\__init__.py", line 525, in as_dict
    ret = meth()
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\__init__.py", line 677, in cmdline
    return self._proc.cmdline()
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\_pswindows.py", line 681, in wrapper
    raise convert_oserror(err, pid=self.pid, name=self._name)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\_pswindows.py", line 671, in convert_oserror
    raise exc
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\_pswindows.py", line 679, in wrapper
    return fun(self, *args, **kwargs)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\_pswindows.py", line 695, in wrapper
    return fun(self, *args, **kwargs)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\psutil\_pswindows.py", line 787, in cmdline
    ret = cext.proc_cmdline(self.pid, use_peb=True)
OSError: [WinError 0] The operation completed successfully: '(originated from OpenProcess)'
```

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -137,7 +137,7 @@ class Job:
         jobs = []
         cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
 
-        for proc in psutil.process_iter(['pid', 'cmdline']):
+        for proc in psutil.process_iter():
             # Ignore processes which most likely have terminated between the time of
             # iteration and data access.
             with contextlib.suppress(psutil.NoSuchProcess, psutil.AccessDenied):


altendky (approved):
Thank you for the detailed explanation and the extremely simple fix.  That's just the way it should be.  `:]`  I read the docs on `psutil.process_iter()` and it does seem to be not needed here.

hackerzhuli:
It did take me some time to come up with this fix. I tried to catch exception on this line for proc in psutil.process_iter(['pid', 'cmdline']), and failed miserably. Then I realized we don't need to get any attributes on this line.

To get plotman to work on Windows,  there are about, 5-7 changes that I need to make. Which I already tested in a private fork. If you'd like I can manually copy thoes changes to my public fork and make pull requests.

There is a catch though, I don't use archiving, so I won't (at least not for now) test whether archving works on Windows.

Here is what I manage to recall(this list is imcomplete):
1. one line added to "setup.cfg":
```
install_requires =
    appdirs
    attrs
    click
    desert
    marshmallow
    pendulum
    psutil ~= 5.8
    pyyaml
    texttable
    windows-curses;platform_system=='Windows' # this is the new line
```
because curses will not run without this line.

2. one change must be made to "interactive.py" to handle the exception in "cusrses_main", these two lines doesn't work on windows:
```
tmp_prefix = os.path.commonpath(cfg.directories.tmp)
dst_prefix = os.path.commonpath(cfg.directories.dst)
```
some exception will be thrown, I don't recall the type, but it seems there is no comman path between some pathes for example:
"C:\Foo" and "D:\Bar"

3. one change must be made to "plotman.py", in method "main":
``` 
if (conf != 'y'):
  print('canceled.  If you wish to resume the job, do so manually.')
else:
  print('killing...')
  job.cancel()
  psutil.wait_procs([job.proc]) # Here is the new line
  print('cleaing up temp files...')
  for f in temp_files:
      os.remove(f)
```
this is added because, job.cancel() does not terminate processes instantatously and causes deleting the temp files to fail (at least on windows) with some sort of file access exception. we must wait for the process to terminate and then delete the temp files.

4. one change must be made to "plot_util.py", in method "df_b", like so:
```
def df_b(d):
    'Return free space for directory (in bytes)'
    #stat = os.statvfs(d)   # this is the original implementation
    total, used, free = shutil.disk_usage(d) # cross platform
    return free
```
https://docs.python.org/3/library/os.html#os.statvfs
as shown in the documentation, this os.statvfs is only available on Unix systems

5. There is one problem with log file paths, that is, like so:
"2021-05-11-07:46:54.log"
colon is not allowed on windows:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

This will also will make git clone fail when checkout files!

But it seems it is already fixed in development branch.

5. one change might need to be made to "plotman.py",  in method "get_term_width":
```
def get_term_width():
    columns = 0
    try:
        (rows, columns) = os.popen('stty size', 'r').read().split()
        columns = int(columns)
    except:
        columns = 120  # 80 is typically too narrow.  TODO: make a command line arg.
    return columns
```
there seems to be a problem, that, the setting "UserInterface.use_stty_size" is not used in this method.
which for obvious reasons needed to be disabled on windows.

altendky:
Deleting files on Windows is spectacularly hard...  Anyways, I do want Windows support so would certainly appreciate the PRs (one for each point, please).  I am also trying to explore options other than curses to make the code more pleasant and more portable.  So maybe worry less about the terminal and curses bits, but definitely submit the rest.  If we get through them and I am not making good progress on the not-curses development, we may merge your curses fixes as well.

Thanks for the further explanation of these additional points.  The help is definitely appreciated.

hackerzhuli:
Okay. Then I will make 1, 3, 4 into PRs, ignore 2, 6, and 5 is already fixed.

PR #361, opened 2021-05-12
fix: make plot_util.df_b (return free space for directory) cross platform (now works on Windows)

hackerzhuli:
https://docs.python.org/3/library/os.html#os.statvfs
as shown in the documentation, "os.statvfs" is only available on Unix systems


altendky:
Hey @hackerzhuli, sorry I haven't found time to look over this yet.  Hopefully I can make some time tomorrow.  Thanks for the help on the Windows support front.

hackerzhuli:
@altendky Don't worry. Take your time.

src/plotman/plot_util.py
```diff
@@ -1,13 +1,14 @@
...
+import shutil
 
 GB = 1_000_000_000
 
 def df_b(d):
     'Return free space for directory (in bytes)'
-    stat = os.statvfs(d)
-    return stat.f_frsize * stat.f_bavail
+    total, used, free = shutil.disk_usage(d)
+    return free
```
altendky, line 11: This is a bit of a nuanced preference against tuples for returning multiple values.  Yeah, the stdlib is unlikely to change this interface ever but in general I would rather use names to identify values rather than order of elements in a sequence.  And yes, we do use order in plenty of places.  It's one of the many things I want to clean up.

Does this look ok to you?

```suggestion
    usage = shutil.disk_usage(d)
    return usage.free
```

altendky (changes requested):
Just to collect my thoughts here...

https://docs.python.org/3.7/library/os.html#os.statvfs
https://docs.python.org/3.7/library/shutil.html#shutil.disk_usage

So, I had a little concern since we were using 'available' space vs. the 'free' wording used by the `shutil.disk_usage()` docs.  But, it does seem to report available, not free.

```console
$ python3 -c 'import os, shutil, sys; arg = sys.argv[1]; stat = os.statvfs(arg); print(stat.f_frsize * stat.f_bfree); print(stat.f_frsize * stat.f_bavail); print(shutil.disk_usage(arg).free)' /
18686353408
16628641792
16628641792
```

So, that's good.

Thanks again.

src/plotman/plot_util.py
hackerzhuli: Okay, who should commit the change.

hackerzhuli: The source code of shutil.disk_usage() does show that `free` is indeed available space in linux systems.

```
if hasattr(os, 'statvfs'):

    __all__.append('disk_usage')
    _ntuple_diskusage = collections.namedtuple('usage', 'total used free')
    _ntuple_diskusage.total.__doc__ = 'Total space in bytes'
    _ntuple_diskusage.used.__doc__ = 'Used space in bytes'
    _ntuple_diskusage.free.__doc__ = 'Free space in bytes'

    def disk_usage(path):
        """Return disk usage statistics about the given path.

        Returned value is a named tuple with attributes 'total', 'used' and
        'free', which are the amount of total, used and free space, in bytes.
        """
        st = os.statvfs(path)
        free = st.f_bavail * st.f_frsize
        total = st.f_blocks * st.f_frsize
        used = (st.f_blocks - st.f_bfree) * st.f_frsize
        return _ntuple_diskusage(total, used, free)

elif _WINDOWS:

    __all__.append('disk_usage')
    _ntuple_diskusage = collections.namedtuple('usage', 'total used free')

    def disk_usage(path):
        """Return disk usage statistics about the given path.

        Returned values is a named tuple with attributes 'total', 'used' and
        'free', which are the amount of total, used and free space, in bytes.
        """
        total, free = nt._getdiskusage(path)
        used = total - free
        return _ntuple_diskusage(total, used, free)
```

altendky: Generally, yes, if you like the suggestion you would click the commit button underneath it.  I just went ahead and did it in this case.

altendky (approved):
I appreciate not only the contributions but your attention to detail as well.  Willingness to dig into the relevant source code even outside are project is certainly valuable.  Cheers.

Issue #374, opened 2021-05-12
Job summary meant for `plot` shows in `interactive`


altendky:
**Describe the bug**
The extra job summary should not be added to the interactive view, I think.  Please consider moving it outside of `plotman.reporting.status_report()` so that it only shows for `plotman plot` output.  I think this was the intent and we both missed the side effect.  If I am misunderstanding something, please let me know.

![image](https://user-images.githubusercontent.com/543719/118047134-5204de00-b348-11eb-874f-fb286eac62bd.png)

**To Reproduce**
Run `plotman interactive` from https://github.com/ericaltendorf/plotman/commit/e055364f5057fda9c9017f1fd0f06e5a29458032

**Expected behavior**
As before, the `Total jobs:` and `Jobs in .:` lines are not present.  I believe this information already has places to be.  The total job count at the top and the tmp section below.  Though, if a separate PR were made to add the integer number rather than just the phase list to the tmp table it might be accepted.  I haven't thought about how well it would fit yet.

altendky:
@rafaelsteil, would you have time to look at this?  I don't think it's particularly urgent.  If not just let me know and I'll get it taken care of.  Thanks.

rafaelsteil:
Yeah, I can take a look!

rafaelsteil:
@altendky Could you elaborate this? `Though, if a separate PR were made to add the integer number rather than just the phase list to the tmp table it might be accepted`. That's the integer number you're referring to?

PR #389, opened 2021-05-13
Removes summary info that was leaking into interactive

rafaelsteil:
A previous PR added a small summary to `reporting.status_report` with the intention to be shown in `plotman status`, however it turns out that `plotman analyze` also uses the same method and the summary was showing there as well. 

Fixes https://github.com/ericaltendorf/plotman/issues/374


altendky (approved):
Thanks for the second pass here.  `:]`

PR #397, opened 2021-05-13
add global stager

wisherhxl:
stager is a great way to manager CPU threads, but only stager for tmp directory is not enough to manage threads efficiently if I have multiple HDDs or SSDs, so I add a global stager, to use with multiple SSDs, and keep ALL threads used by running ploting won't exceed thread number of the computer.


altendky:
If you want to limit overall threads, how about we talk about a global thread limit?  I had suggested this a few days ago.  Feel free to open up a ticket.

wisherhxl:
that would be nice.

Issue #405, opened 2021-05-13
plotman tracking same plot multiple times


acceleratxr:
**Describe the bug**

Plotman is tracking each plot twice.

**To Reproduce**
1. Install current plotman `pip install --force-reinstall git+https://github.com/ericaltendorf/plotman@development`.
2. Run `plotman interactive`

**Expected behavior**

Each listed job has a unique plot id.

**System setup:**
 - OS: Ubuntu 20.04 LTS

**Config**
<details> <summary>full configuration</summary>

```yaml
# Options for display and rendering
user_interface:
        # Call out to the `stty` program to determine terminal size, instead of
        # relying on what is reported by the curses library.   In some cases,
        # the curses library fails to update on SIGWINCH signals.  If the
        # `plotman interactive` curses interface does not properly adjust when
        # you resize the terminal window, you can try setting this to True.
        use_stty_size: True

# Where to plot and log.
directories:
        # One directory in which to store all plot job logs (the STDOUT/
        # STDERR of all plot jobs).  In order to monitor progress, plotman
        # reads these logs on a regular basis, so using a fast drive is
        # recommended.
        log: /var/log/chia

        # One or more directories to use as tmp dirs for plotting.  The
        # scheduler will use all of them and distribute jobs among them.
        # It assumes that IO is independent for each one (i.e., that each
        # one is on a different physical device).
        #
        # If multiple directories share a common prefix, reports will
        # abbreviate and show just the uniquely identifying suffix.
        tmp:
                - /chia/plotter0
                - /chia/plotter1
                - /chia/plotter2

        # Optional: Allows overriding some characteristics of certain tmp
        # directories. This contains a map of tmp directory names to
        # attributes. If a tmp directory and attribute is not listed here,
        # it uses the default attribute setting from the main configuration.
        #
        # Currently support override parameters:
        #     - tmpdir_max_jobs
        tmp_overrides:
                # In this example, /mnt/tmp/00 is larger than the other tmp
                # dirs and it can hold more plots than the default.
                "/chia/plotter2":
                        tmpdir_max_jobs: 6

        # Optional: tmp2 directory.  If specified, will be passed to
        # chia plots create as -2.  Only one tmp2 directory is supported.
        # tmp2: /mnt/tmp/a

        # One or more directories; the scheduler will use all of them.
        # These again are presumed to be on independent physical devices,
        # so writes (plot jobs) and reads (archivals) can be scheduled
        # to minimize IO contention.
        dst:
[...]

Issue #337, opened 2021-05-11
Double count of jobs


D4rk4:
> 
> $ plotman status
>  plot id    k              tmp              dst   wall   phase    tmp       pid   stat     mem   user    sys     io
> fc950a1b   32   /srv/chia/fast   /srv/chia/dest   6:58     3:1   236G   4176527    SLP    2.6M     0s     0s     0s
> fc950a1b   32   /srv/chia/fast   /srv/chia/dest   6:58     3:1   236G   4176528    DSK   18.4G   6:01   0:07   1:23
> a9f3ea7a   32   /srv/chia/fast   /srv/chia/dest   9:40     4:0   176G   1317240    SLP    2.6M     0s     0s     0s
> a9f3ea7a   32   /srv/chia/fast   /srv/chia/dest   9:40     4:0   176G   1317241    RUN   17.7G   8:16   0:13   4:19

Chia from deb package:

> $ dpkg -l | grep chia; which chia; ls -lash `which chia`
> ii  chia-blockchain                               1.1.4                                amd64        GUI for Chia Blockchain
> /usr/local/bin/chia
> 512 lrwxrwxrwx 1 root staff 64 мая  8 17:57 /usr/local/bin/chia -> /usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon/chia

altendky:
Thanks.  I'll try to figure this out tonight.  I do see #330 but would like to understand this a bit better.  I'm a bit confused why this is just popping up now.

juliencampy:
I got a fix on my computer but I won't create a pull request for this as this is a hack at best.
It is adding this filter in the get_running_jobs function :

`proc.cpu_times().user == 0.0`

It looks counter intuitive, but only the proc with this attributes at 0 does have open_files associated with it. I am not a linux nor python expert so I have no idea why ...

It works here but I guess that will not be the case on computer not affected by the issue.
A cleaner way would be to use the ppid() of the proc, but I can't make it work atm.

Cheers.

lexzz:
I'm seeing the same issue on the development branch

rafaelsteil:
> I got a fix on my computer but I won't create a pull request for this as this is a hack at best.
> It is adding this filter in the get_running_jobs function :
> 
> `proc.cpu_times().user == 0.0`
> 
> It looks counter intuitive, but only the proc with this attributes at 0.0 does have open_files associated with it. I am not a linux nor python expert so I have no idea why ...
> 

[...]

Issue #301, opened 2021-05-10
Install Plotman w/ Chia's Experimental GUI Installer


ijf4ghub313:
Linux Mint 20.04
I installed Chia via the "experimental GUI installer" (_chia-blockchain_1.1.5_amd64.deb_) and because it doesn't quite operate the same ---no requirement to . ./activate the virtual environment, etc--- I'm not sure if I'm performing the Plotman install correctly (actually I'm pretty sure I'm not). 

Any insight on the differences of installing Plotman if I used Chia's "experimental GUI installer"? 
( The Chia GUI does seem to be operating fine ... I'm synced and generating plots like mad. )

But I'm not even sure where Chia installed to ... looks like it might be in `/usr/lib/chia-blockchain` directory(?), but . ./activate doesnt work from there and when I try to install Plotman i get a 'launchpadlib' error and a "`WARNING: The script plotman is installed in '/home/ijf/.local/bin' which is not on PATH.`" during install. And nothing seems to work after install either ... "`plotman: command not found`".

I think if I could just get over this hump your instructions are very followable.

bradsquicciarini:
If you installed it with the experimental GUI installer, it won't work. It can't find active jobs. I have an open pull request to allow it to work with the experimental GUI install. 

But as for your install problems, you just need to sym link it.

```
# Find the location of plotman
pip show plotman

# Use the path from that output to sym link it to something in your $PATH
ln -s <path to plotman> /usr/local/bin/plotman
```

altendky:
Did you try the development branch?  There are various fixes related to not-Python installs of chia.

ijf313-github:
Thanks for the help guys. I think I'll just wait for a future version. I'm new enough with Linux, Chia, and GitHub that I'll just be wasting your time asking noob questions. Looking forward to using Plotman in the future. Thanks for the work you're doing. -ijf

altendky:
FWIW, installing the development version of plotman involves following the regular installation instructions and using `@development` rather than `@main`.  But sure, I might get a release out tonight, maybe.

lexzz:
Hey! I'm having the same question.

Tried to install the development version but seeing "plotman: command not found" 
As described I made a symlink but maybe forgot something? Hope you can give me a hint.

Thanks in advance!

lexzz:
@ijf313-github Thanks for your message

After some trying I was able to install Plotman with pip3 but still wasn't able to get it going.
[...]

PR #418, opened 2021-05-14
fix: skip so-called plot processes that are parents of others

altendky:
The new experimental Chia GUI installer launches plot processes
such that you end up with two processes with the exact same
command line.  This confuses us since we identify both as being
separate plot processes and report two when there's only one that
is actually interesting.

This fix identifies the parents of all the potential plotting
processes and removes them from the list of processes of interest.

https://github.com/ericaltendorf/plotman/issues/301
https://github.com/ericaltendorf/plotman/pull/330
https://github.com/ericaltendorf/plotman/issues/337
https://github.com/ericaltendorf/plotman/issues/405

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -137,28 +137,46 @@ class Job:
         jobs = []
         cached_jobs_by_pid = { j.proc.pid: j for j in cached_jobs }
 
-        for proc in psutil.process_iter():
-            # Ignore processes which most likely have terminated between the time of
-            # iteration and data access.
-            with contextlib.suppress(psutil.NoSuchProcess, psutil.AccessDenied):
-                if is_plotting_cmdline(proc.cmdline()):
-                    if proc.pid in cached_jobs_by_pid.keys():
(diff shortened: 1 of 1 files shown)


src/plotman/job.py
```diff
@@ -137,28 +137,45 @@ def get_running_jobs(logroot, cached_jobs=()):
...
+
+            pids = {process.pid for process in processes}
+            ppids = {process.ppid() for process in processes}
+            wanted_pids = pids - ppids
+
+            wanted_processes = [
+                process
+                for process in processes
+                if process.pid in wanted_pids
+            ]
```
rafaelsteil, line 159: A nit but a good nit: add a small comment saying why this is necessary, maybe even reference this PR, as it may not be clear the reason this is necessary

altendky: Thanks, added.  Does it read well?

rafaelsteil: Superb 👨‍🍳 💋

altendky:
Got some user feedback this works.

PR #427, opened 2021-05-14
tmp_overrides add some attr

xiaobear250:
I add  
   tmpdir_stagger_phase_major: 2
   tmpdir_stagger_phase_minor: 1
   tmpdir_stagger_phase_limit: 2
in config - tmp_overrides.
I think it is usefull when a tmp larger than the other tmp.

I had try in manager.py 60-64:
tmp_cfg=sched_cfg.copy()
tmp_cfg.update(curr_overrides)
and then use tmp_cfg.tmpdir_max_jobs

but sched_cfg is class,not dict.


adrienbrault:
This feature would be useful to better optimize settings when using mixed tmp drives, for example a mix of fast NVMe and slower SATA drives.

altendky:
Thank you for your effort contributing this PR to plotman.  I apologize for failing to keep up with the community contributions.  It looks like this is very similar to https://github.com/ericaltendorf/plotman/pull/758.  It is kind of a toss up, but I expect to work through https://github.com/ericaltendorf/plotman/pull/758 instead and hopefully get it merged in the near future.  Please do let me know if there is anything that it is missing.

PR #455, opened 2021-05-15
adds -x parameter to plot command

mruckli:
Introduces the possibility to skip adding the dst / final_dir
to the harvester for farming.
Especially useful if you have harvesters that are running somewhere else
and you are just plotting on the machine where plotman is running.

### DIFF
--- a/src/plotman/configuration.py
+++ b/src/plotman/configuration.py
@@ -89,6 +89,7 @@ class Plotting:
     job_buffer: int
     farmer_pk: Optional[str] = None
     pool_pk: Optional[str] = None
+    x: bool = False
 
 @attr.frozen
 class UserInterface:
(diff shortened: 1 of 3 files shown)


src/plotman/configuration.py
```diff
@@ -89,6 +89,7 @@ class Plotting:
     job_buffer: int
     farmer_pk: Optional[str] = None
     pool_pk: Optional[str] = None
+    x: Optional[bool] = None
```
altendky, line 92: We don't actually need this to be optional I don't think, just a default value cuts it.

```suggestion
    x: bool = False
```

altendky (changes requested):
Yeah, this seems a good thing to have.  Thanks for the contribution.  `:]`

One point about the change noted inline and also this needs to be targeted at development.  You should be able to click the `Edit` button to the right of the PR title to change it.  Probably need an update merge from development, button down below for that.  Or I can do all this bookkeeping, but I try not to butt in on other people's PRs.

mruckli:
Rebased onto development and integrated your suggestion.

altendky (approved):
Thanks again for the clean and clear contribution.  I'll likely be enabling this on my own systems soon.

PR #471, opened 2021-05-15
calculate plot size given the parameter k

hackerzhuli:
Using a formula instead of a constant is better for improving plotman to work with plots with a different k.

Formula from [Chia Proof of Space Construction](https://www.chia.net/assets/proof_of_space.pdf) "Space Required" section.

image

### DIFF
--- a/src/plotman/_tests/plot_util_test.py
+++ b/src/plotman/_tests/plot_util_test.py
@@ -57,3 +57,10 @@ def test_list_k32_plots(fs: pyfakefs.fake_filesystem.FakeFilesystem):
             [ '/t/plot-k32-0.plot',
               '/t/plot-k32-1.plot',
               '/t/plot-k32-5.plot' ] )
+
+
+def test_get_plotsize():
+    assert (
+        [659272492, 107287518791, 221143636517, 455373353413, 936816632588]
+        == [plot_util.get_plotsize(n) for n in [25, 32, 33, 34, 35]]
+    )
--- a/src/plotman/chiapos.py
+++ b/src/plotman/chiapos.py
@@ -1,0 +1,87 @@
+# version = 1.0.2
+# https://github.com/Chia-Network/chiapos/blob/1.0.2/LICENSE
+# https://github.com/Chia-Network/chiapos/blob/1.0.2/src/pos_constants.hpp
+# start ported code
+# Unique plot id which will be used as a ChaCha8 key, and determines the PoSpace.
+kIdLen = 32;
+
+# Distance between matching entries is stored in the offset
+kOffsetSize = 10;
+
+# Max matches a single entry can have, used for hardcoded memory allocation
+kMaxMatchesSingleEntry = 30;
+kMinBuckets = 16;
+kMaxBuckets = 128;
+
+# During backprop and compress, the write pointer is ahead of the read pointer
+# Note that the large the offset, the higher these values must be
+kReadMinusWrite = 1 << kOffsetSize;
+kCachedPositionsSize = kReadMinusWrite * 4;
+
+# Must be set high enough to prevent attacks of fast plotting
+kMinPlotSize = 18;
+
+# Set to 50 since k + kExtraBits + k*4 must not exceed 256 (BLAKE3 output size)
+kMaxPlotSize = 50;
+
+# The amount of spare space used for sort on disk (multiplied time memory buffer size)
+kSpareMultiplier = 5;
+
+# The proportion of memory to allocate to the Sort Manager for reading in buckets and sorting them
+# The lower this number, the more memory must be provided by the caller. However, lowering the
+# number also allows a higher proportion for writing, which reduces seeks for HDD.
+kMemSortProportion = 0.75;
+kMemSortProportionLinePoint = 0.85;
+
+# How many f7s per C1 entry, and how many C1 entries per C2 entry
+kCheckpoint1Interval = 10000;
+kCheckpoint2Interval = 10000;
+
+# F1 evaluations are done in batches of 2^kBatchSizes
+kBatchSizes = 8;
+
+# EPP for the final file, the higher this is, the less variability, and lower delta
+# Note: if this is increased, ParkVector size must increase
+kEntriesPerPark = 2048;
+
+# To store deltas for EPP entries, the average delta must be less than this number of bits
+kMaxAverageDeltaTable1 = 5.6;
+kMaxAverageDelta = 3.5;
+
+# C3 entries contain deltas for f7 values, the max average size is the following
+kC3BitsPerEntry = 2.4;
+
+# The number of bits in the stub is k minus this value
+kStubMinusBits = 3;
+
+#end ported code
+
+# version = 1.0.2
+# https://github.com/Chia-Network/chiapos/blob/1.0.2/LICENSE
+# https://github.com/Chia-Network/chiapos/blob/1.0.2/src/util.hpp
+# start ported code
+def ByteAlign(num_bits):
+    return (num_bits + (8 - ((num_bits) % 8)) % 8)
+# end ported code
+
+# version = 1.0.2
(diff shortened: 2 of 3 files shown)


hackerzhuli:
Test failed because the test file size 100GB is too close to rule out that is not a k32 plot.

altendky (changes requested):
I think perhaps that document is out of date.  All my plots are in the 108.8GB range rather than 104.7GB.  Do you have time to dig into this discrepancy?

```console
$ ls -l /farm/sites/007 | tail -n 5
-rw-r--r-- 1 chia chia 108861233015 May 15 05:30 plot-k32-2021-05-14-13-46-8b913363152db776f47dd84a73b28c213c479ad9eb6cf9c5b62f5e53b260f945.plot
-rw-r--r-- 1 chia chia 108833740950 May 15 07:50 plot-k32-2021-05-14-16-07-8f281deb8a61e7d8240a3c023b40fdc4ae37d30221af1e18a36a1248da629740.plot
-rw-r--r-- 1 chia chia 108824815844 May 15 09:30 plot-k32-2021-05-14-17-48-ad72c2d71419d10b4ae4ecc13d8e453ab154f77c7a07d254a47e781fb3519b7c.plot
-rw-r--r-- 1 chia chia 108790088102 May 15 11:18 plot-k32-2021-05-14-19-31-b981549c5e8bd3f26b1c47b10c400b42e79ec79dc3d7c40ea41ef223297923f3.plot
-rw-r--r-- 1 chia chia 108884412258 May 15 13:09 plot-k32-2021-05-14-21-15-8fec2777aa2cc7097fd52ea4ce7d70c3f6f494e3b55688ef914fa8574477c700.plot
```
```console
$ python -c 'print(int(0.762 * 32 * 2**32))'
104728482545
```

hackerzhuli:
I'm not sure where the diffrence come from. I'll take a look into it.

hackerzhuli:
The formula in the document, assumes that each entry on average takes k bits.
But the entries are "parked", that is, each 2048 entries are stored together as a group.
But the actual size needed for each park is different.
Chia used constant park size to increase performance.
So, there are spaces wasted in the parks.

By looking into the source code of chiapos, I am able to find exactly how park size is calculated and thus account for some of the space differences.

I wrote code to find out the plotsize taking park sizes into account. plot_size.py:
```
# the following constants are from chiapos pos_constants.hpp

# EPP for the final file, the higher this is, the less variability, and lower delta
# Note: if this is increased, ParkVector size must increase
kEntriesPerPark = 2048

# To store deltas for EPP entries, the average delta must be less than this number of bits
kMaxAverageDeltaTable1 = 5.6;
kMaxAverageDelta = 3.5

# The number of bits in the stub is k minus this value
kStubMinusBits = 3

def calc_average_size_of_entry(k, table_index):
    '''
    calculate the average size of entries in bytes
    the average size of entry is approximately k/8 bytes
    but due to wasted space of parks, it is larger than that
    '''
    return calc_park_size(k, table_index) / kEntriesPerPark

def byte_align(num_bits):
    return (num_bits + (8 - ((num_bits) % 8)) % 8)

def calc_park_size(k, table_index):
    '''
    park size in bytes for storing kEntriesPerPark = 2048 entries
    it is approximately kEntriesPerPark * k / 8, but it is actually larger than that
    derived from chiapos EntrySizes::CalculateParkSize
    '''
    line_point_size_bits = byte_align(2 * k)
    stub_size_bits = byte_align((kEntriesPerPark - 1) * (k - kStubMinusBits))
    max_delta_size_bits = byte_align((kEntriesPerPark - 1) * kMaxAverageDeltaTable1 if table_index == 1 else (kEntriesPerPark - 1) * kMaxAverageDelta)
    return (line_point_size_bits + stub_size_bits + max_delta_size_bits) / 8

def get_probability_of_entries_kept(k, table_index):
    '''
    get the probibility of entries in table of table_index that is not dropped
    the formula is derived from https://www.chia.net/assets/proof_of_space.pdf,  section Space Required, p5 and pt
    '''

    if table_index > 5:
        return 1

    power = 2**k
    
    if table_index == 5:
        return 1 - (1 - 2 / power) ** power    # derived from Space Required p5
    else:
        return 1 - (1 - 2 / power) ** (get_probability_of_entries_kept(k, table_index + 1) * power) # derived from Space Required pt

def get_plotsize_scaler(k):
    '''
    get scaler for plot size so that the plot size can be calculated by scaler * k * 2 ** k
    '''

    result = 0
    for i in range(1, 8):
        probability = get_probability_of_entries_kept(k, i)
        average_size_of_entry = calc_average_size_of_entry(k, i)
        scaler_for_table = probability * average_size_of_entry / k
        #print(f"probability = {probability}, size = {average_size_of_entry}, scaler = {scaler_for_table}")
        result += scaler_for_table

    return result    
```


console output:
```
k = 20, scaler = 0.792, size = 0.02 GB
k = 21, scaler = 0.790, size = 0.03 GB
k = 22, scaler = 0.789, size = 0.07 GB
k = 23, scaler = 0.788, size = 0.15 GB
k = 24, scaler = 0.787, size = 0.32 GB
k = 25, scaler = 0.786, size = 0.66 GB
k = 26, scaler = 0.785, size = 1.37 GB
k = 27, scaler = 0.784, size = 2.84 GB
k = 28, scaler = 0.783, size = 5.89 GB
k = 29, scaler = 0.783, size = 12.18 GB
k = 30, scaler = 0.782, size = 25.19 GB
k = 31, scaler = 0.781, size = 52.01 GB
k = 32, scaler = 0.781, size = 107.29 GB
k = 33, scaler = 0.780, size = 221.14 GB
k = 34, scaler = 0.780, size = 455.37 GB
k = 35, scaler = 0.779, size = 936.82 GB
k = 36, scaler = 0.779, size = 1925.98 GB
k = 37, scaler = 0.778, size = 3957.05 GB
k = 38, scaler = 0.778, size = 8123.48 GB
k = 39, scaler = 0.777, size = 16665.72 GB
```

This is indeed much closer than 0.762. It turns out the scaler is different for different k, probabally due to space waste differences.

altendky:
Is there any of the Chia code that would be useful to copy and paste?  We already to that for the command line parsing of plotting processes so we could do it for other bits as needed.  Also, there is a mechanism to track the different versions of chia, though they haven't changed that code since we started copying it.

https://github.com/ericaltendorf/plotman/blob/1876d4cbda73fb38584f95113a00f01af1e21704/src/plotman/chia.py

altendky:
Oh, and thanks for the hard work here.  `:]`

hackerzhuli:
We cannot copy the code in this case, because chiapos is implemented in C++. I changed them a bit to run in python. Also, I think the code is too complex for this simple usage.

altendky:
I would default to having the actual calculations present in the code.  If they are kind of slow then we can run them as needed, or at start up, and cache results.  If they are super slow then we can consider pre-generation.

I think it is reasonable to continue these here and view this PR as a general improvement on the assumptions and simplifications that plotman has at present.  It would be good to provide pinned links in the code to the C++ code it 'mirrors'.  This allows for easier maintenance and verification.  Let's pin the links to release versions as in the link I shared for the copied code.

I'm thinking I may restructure that to have a separate module per version like `plotman/chia/v1_1_5.py` or such and an interface to get at each by just latest version or a specific version.  But, I can follow up with that when I have time.  In case I hadn't mentioned it to you, I forget where I wrote it, I imagine that we could identify the version of chia that each plotting process is using and use the proper cli parser etc for it.  Someday...

hackerzhuli:
I moved to code that is from chiapos to chiapos.py, since it is different from chia-blockchain.

I think code ported from chiapos should not be duplicated every time there is a new version, since the code from chiapos is much less likely to change.

hackerzhuli:
I just realized that the size calculated here is the minimal size of the plot file. So we expect an actual file will be about 1-2% larger than it(I guess, there is no proof though). This can be used to tell whether a plot file is of certain size K and whether it is complete.

altendky (approved):

PR #486, opened 2021-05-16
k32 plot sizes approach 109GB

nbaird:
k32 plot sizes are underestimated at 108GB


altendky:
This is being worked on in more detail over in https://github.com/ericaltendorf/plotman/pull/471.  Will the solution we are working there be sufficient?

nbaird:
Yes I believe so. I will close this.

PR #509, opened 2021-05-17
adds logs printing command

mruckli:
Introduces a logs command inspired by how `docker logs` works when interacting with containers but now for plots via plotman.
Basically I got sick of running `plotman status` then `plotman details` and afterwards `tail` or `less` for the given logfile. Now it is possible to simply print or follow the logs after getting the id of the plot via `plotman status`.

### DIFF
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -413,7 +413,7 @@ class Job:
             logfile = self.logfile
             )
 
-    def print_logs(self, follow=None):
+    def print_logs(self, follow: bool = False) -> None:
         with open(self.logfile, 'r') as f:
             if follow:
                 line = ''
--- a/src/plotman/job.py
+++ b/src/plotman/job.py
@@ -412,7 +412,23 @@ class Job:
             dst = self.dstdir,
             logfile = self.logfile
             )
-    
+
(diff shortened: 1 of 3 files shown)


altendky:
I apologize for not keeping up with the community contributions.  This does look like a nice, concise, and useful addition.  Do you have time to catch it up with the changes to the `development` branch?  If so I can take a look and hopefully get it merged soon.

mruckli:
No worries, I understand that it must be hectic with so many things to work on :)
I updated it to the latest development branch.

src/plotman/job.py
```diff
@@ -412,7 +412,23 @@ def status_str_long(self) -> str:
             dst = self.dstdir,
             logfile = self.logfile
             )
-    
+
+    def print_logs(self, follow=None):
```
altendky, line 416: I think this might be enough to satisfy the type hints checking.

```suggestion
    def print_logs(self, follow: bool = False) -> None:
```

altendky (changes requested):

altendky:
I hope you don't mind me butting in and finishing out the remaining boilerplate.  Thanks for your contribution here.  `:]`

PR #532, opened 2021-05-17
Add a log parser class

rafaelsteil:
Adds a new log parser for finished plotting jobs and a data structure for all the information found. The intention is to provide a decoupled and centralized class that can be used for any other command that relies on plot log information, like analize, a future csv exporter, graph generation etc.


src/plotman/plotinfo.py
```diff
@@ -0,0 +1,108 @@
...
+    tmp_dir2: str = ""
+    phase1_duration_raw: float = 0
+    phase2_duration_raw: float = 0
+    phase3_duration_raw: float = 0
+    phase4_duration_raw: float = 0
+    total_time_raw: float = 0
+    copy_time_raw: float = 0
+    filename: str = ""
+
+    def is_empty(self) -> bool:
```
altendky, line 22: Wouldn't this indicate it being incomplete, not empty?

```suggestion
    def is_incomplete(self) -> bool:
```

src/plotman/log_parser.py
```diff
@@ -0,0 +1,140 @@
...
+        m = re.search(r'^Plot size is: (\d+)', line)
+        if m:
+            entry.plot_size = int(m.group(1))
+        return m != None
+
+    # Starting phase 1/4: Forward Propagation into tmp files... Sun May  9 17:36:12 2021
+    def plot_start_date(self, line: str, entry: PlotInfo) -> bool:
+        m = re.search(r'^Starting phase 1/4: Forward Propagation into tmp files\.\.\. (.+)', line)
+        if m:
+            entry.started_at = m.group(1)
```
altendky, line 133: Couldn't we parse this here to be a `pendulum.datetime.DateTime` or such?  If so, note the type hint and default below as well.

https://github.com/ericaltendorf/plotman/blob/7df92df56ea6e99579bfbb83e4a55687c1380a39/src/plotman/job.py#L41-L43

src/plotman/plotinfo.py
```diff
@@ -0,0 +1,108 @@
+import attr
+
+@attr.s(repr=True, init=False)
```
altendky, line 3: Why no `.__init__()`?  Maybe `order=False` though.  I would have expected that `repr` defaults to effectively being true?

src/plotman/log_parser.py
```diff
@@ -0,0 +1,140 @@
+import re
+from plotman.plotinfo import PlotInfo
+
+class PlotLogParser:
+    """Parser for a finished plotting job"""
+
+    def parse(self, filename: str) -> PlotInfo:
```
altendky, line 7: How about taking an open file object?  Makes this more flexible and encourages explicit IO control at the top level.

src/plotman/_tests/log_parser_test.py
```diff
@@ -0,0 +1,54 @@
...
+from plotman.log_parser import PlotLogParser
+
+def test_should_correctly_parse():
+  with importlib.resources.path(resources, "2021-04-04T19_00_47.681088-0400.log") as p:
+    logfile_path = p
+
+  parser = PlotLogParser()
+  info = parser.parse(logfile_path)
+
+  assert info.plot_id == "3eb8a37981de1cc76187a36ed947ab4307943cf92967a7e166841186c7899e24"
```
altendky, line 13: Maybe create a reference `PlotInfo` instance and assert equality to that?  That guarantees the complete object is compared whereas here it is easy to leave an attribute unchecked.

altendky (changes requested):
I'll offer to do the leg work on changes here, if you prefer.  You should have gotten a complete review on this long ago.  My apologies for the wait both for the review and inclusion of this in plotman proper.

src/plotman/log_parser.py
```diff
@@ -0,0 +1,140 @@
+import re
+from plotman.plotinfo import PlotInfo
+
+class PlotLogParser:
```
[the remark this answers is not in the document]
altendky: Alternatively, maybe it would make sense to hold onto these objects with them holding the open files and then we can incrementally parse the logs instead of reparsing them every loop, once this is used for the main job info log parsing.  At least I think that's what is done now, I didn't check back right now...

rafaelsteil:
_(Discussed the comments on Keybase)_

PR #550, opened 2021-05-18
Update archive.py

alillebakk:
Removed a / who made disk space on archive not functioning.
Corrected the rsync url so it do not look for  12000 folder and removed rsync://  to get it to work at the different locations I tested the code.

### DIFF
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -86,7 +86,7 @@ def compute_priority(phase, gb_free, n_plots):
 
 def get_archdir_freebytes(arch_cfg):
     archdir_freebytes = {}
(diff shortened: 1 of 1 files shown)


altendky:
While this may help your case, it breaks other cases.  See https://github.com/ericaltendorf/plotman/pull/627 for a rework of the archival system that allows customization.  Maybe we can merge it in the next few days...  Hopefully...

altendky:
If the new archival approach isn't flexible enough for your needs then let's discuss opportunities to improve it.  At this point this PR is not applicable since the code has been significantly reworked.  Thanks for your effort on the project and my apologies for my slowness in handling contributions.

PR #557, opened 2021-05-18
Add command to export log files to CSV

rafaelsteil:
This PR depends on https://github.com/ericaltendorf/plotman/pull/532

Adds a command to export all plot logs as CSV. It can be to STDOUT (default) or to a specific file via the flag `-o <filename>`


src/plotman/plotman.py
```diff
@@ -159,8 +163,15 @@ def main():
...
+    # 
+    # Exports log metadata to CSV
+    #
+    elif args.cmd == 'export_csv':
+        logfilenames = glob.glob(os.path.join(cfg.directories.log, '*'))
+        csv_exporter.export(logfilenames, args.save_to)
+
     else:
-        jobs = Job.get_running_jobs(cfg.directories.log)
+        jobs = Job.get_running_jobs(cfg.directories.log())
```
altendky, line 174: I think?

```suggestion
        jobs = Job.get_running_jobs(cfg.directories.log)
```

src/plotman/plotman.py
```diff
@@ -159,8 +163,15 @@ def main():
         analyzer.analyze(args.logfile, args.clipterminals,
                 args.bytmp, args.bybitfield)
 
+    # 
+    # Exports log metadata to CSV
+    #
+    elif args.cmd == 'export_csv':
```
altendky, line 169: What do you think of it just being an export and the presently implemented format is csv?

src/plotman/csv_exporter.py
```diff
@@ -0,0 +1,115 @@
...
+def generate(logfilenames, out):
+    writer = csv.writer(out)
+    header(writer)
+    logs = parse_logs(logfilenames)
+
+    for info in logs:
+        writer.writerow([
+            info.plot_id,
+            info.started_at,
+            parse_date(info.started_at).strftime("%d %b %Y"),
```
altendky, line 83: Seems like a well formed standard like ISO 8601 or RFC 3339 would be better than a locale specific time here?

src/plotman/csv_exporter.py
```diff
@@ -0,0 +1,115 @@
...
+        save_to_file(logfilenames, save_to)
+
+def save_to_file(logfilenames, filename: str):
+    with open(filename, 'w') as file:
+        generate(logfilenames, file)
+
+def send_to_stdout(logfilenames):
+    generate(logfilenames, sys.stdout)
+
+def header(writer):
```
altendky, line 20: This is a pretty long list to just carefully keep aligned between the two places that the columns are and values are identified.  https://github.com/ericaltendorf/plotman/pull/299/files#diff-62b25901762e697ab64f8088b67b720707e694198e9d26ce3fafd3d08b14006fR262-R325 or such maybe but let's take a look again when the other files are here.

altendky (changes requested):
I think we are missing `plotman.log_parser` and `plotman.plotinfo`?

rafaelsteil:
> I think we are missing `plotman.log_parser` and `plotman.plotinfo`?

This PR depends on https://github.com/ericaltendorf/plotman/pull/532

src/plotman/plotman.py
rafaelsteil: Sounds good to me

src/plotman/csv_exporter.py
rafaelsteil: Yeah, makes sense. I"ll adjust

src/plotman/csv_exporter.py
rafaelsteil: Not sure if I understood. Mind to clarify?

src/plotman/csv_exporter.py
```diff
@@ -0,0 +1,115 @@
...
+def generate(logfilenames, out):
+    writer = csv.writer(out)
+    header(writer)
+    logs = parse_logs(logfilenames)
+
+    for info in logs:
+        writer.writerow([
+            info.plot_id,
+            info.started_at,
+            parse_date(info.started_at).strftime('%Y-%m-%d'),
```
rafaelsteil, line 83: `iso` without time

src/plotman/csv_exporter.py
altendky: There are 34 items here and also 34 items specified around line 80.  It would be easy to get these out of order, or add a new item one place and not the other.  There would be no exception, I don't think, just misaligned column headings and data.  The linked code certainly isn't the only solution.  It was just one that I had coded recently (at that point).  Looking here again now I see that many of the fields are calculated from others.  This makes the linked code less obviously good.

Another thought, have a function that can create a `dict` from a `PlotInfo` object, maybe a method on `PlotInfo`, and use the keys from that as the header.  There is also a [`.writeheader()` method](https://docs.python.org/3/library/csv.html#csv.DictWriter.writeheader).  Then when you pass a dict to be written it will make sure the data goes in the matching column.

src/plotman/csv_exporter.py
```diff
@@ -0,0 +1,115 @@
...
+from plotman.plotinfo import PlotInfo
+
+def export(logfilenames, save_to = None):
+    if save_to is None:
+        send_to_stdout(logfilenames)
+    else:
+        save_to_file(logfilenames, save_to)
+
+def save_to_file(logfilenames, filename: str):
+    with open(filename, 'w') as file:
```
altendky, line 14: https://docs.python.org/3/library/csv.html#examples

If I remember correctly, this relates to quoted fields with newlines in them.  Unlikely, but best to just always open files as the CSV modules suggests.  (detail, py2 is different...)

```suggestion
    with open(filename, 'w', newline='') as file:
```

src/plotman/csv_exporter.py
```diff
@@ -0,0 +1,115 @@
+import csv
+import sys
+from dateutil.parser import parse as parse_date
+from plotman.log_parser import PlotLogParser
+from plotman.plotinfo import PlotInfo
+
+def export(logfilenames, save_to = None):
```
altendky, line 7: I would probably tend to leave the file opening to the caller instead of having these small 'helpers'.  It just generally seems better to leave the file handling explicit at the higher levels.

altendky (changes requested):
Likewise here on the offer to take up the leg work from here.  Thanks for your contributions here and elsewhere.  It's very helpful to have someone else involved in the code.

rafaelsteil:
_(Discussed the comments on Keybase)_

PR #612, opened 2021-05-21
Analysis: added matplotlib figure generation with several desirable performance metrics. 

nikwl:
Thanks so much for this awesome tool!

I've been wanting to see a more graphical representation of how my plots are doing. This pr adds the option to create a matplotlib figure that summarizes your plotting performance, including:
- Dumbbell plot visualizing start, end, and duration of your plots over time. 
- Average plot rate over time (computed over a sliding window). 
- Average plot time (computed over a sliding window). 
- Cumulative number of plots over time.

plotman_figure

It adds two new arguments to ```plotman analyze```:
```--logdir```: a directory that contains a number of log files (this can be used instead --logfile). 
```--figfile```: path to save the figure to. 

It adds one new dependency: ```matplotlib```

The figure creation is done by high level functions which are easily removable or interchangeable. If the maintainers don't like my method of integration, figure creation should be easy to reintegrate in a better way. Or if using matplotlib is undesirable then hopefully the body of the functions can be reused in some way to still be helpful for doing analysis. 

Thanks!

### CODE REVIEW

**Comment by rafaelsteil** [comment]
Here's a little script I created that generated a PDF with charts using data from `plotman export` https://github.com/rafaelsteil/plotman/commit/e4a063220e34d4cc86081b2abb8ff1c02a9aa2e3

File: setup.cfg [hunk 1]
```diff
@@ -45,6 +45,7 @@ install_requires =
     psutil ~= 5.8
     pyyaml ~= 5.4
     texttable ~= 1.6
+    matplotlib ~= 3.4.2
```
altendky, line 48: There should probably be an extra down below for ~~plotting~~ graphing.

```ini
graph =
    matplotlib ~= 3.4
    numpy ~= 1.20
```

altendky (commented):
Thanks for your contribution here.  I apologize for leaving it hanging for so long.

What do you think of separating it from analyze and making it its own `plotman graph` subcommand or such?  I think it could get its own `plotman.graph` module as well.

src/plotman/graph.py
```diff
@@ -0,0 +1,263 @@
...
+            # Record of slicing and data associated with the slice
+            sl = 'x'         # Slice key
+            phase_time = {}  # Map from phase index to time
+            n_sorts = 0
+            n_uniform = 0
+            is_first_last = False
+
+            # Read the logfile, triggering various behaviors on various
+            # regex matches.
+            for line in f:
```
altendky, line 138: Do we need dedicated log parsing here?  If the existing parsing doesn't catch everything, maybe we could add the extra bits there?  Though I have to admit we have two parsers right now.  I was hoping to make some time to get everything using the newer one.  I'll try to get that done first.

src/plotman/plotman.py
```diff
@@ -94,8 +94,24 @@ def parse_args(self) -> typing.Any:
         p_analyze.add_argument('--bybitfield',
                 action='store_true',
                 help='slice by bitfield/non-bitfield sorting')
-        p_analyze.add_argument('logfile', type=str, nargs='+',
+        p_analyze.add_argument('--logfile', type=str, nargs='+', default=None,
                 help='logfile(s) to analyze')
+        p_analyze.add_argument('--logdir', type=str, default=None,
+                help='directory containing multiple logfiles to analyze')
```
altendky, line 100: This and the same below in `graph` seem to be unused.  I think if this is removed then the `--` can be removed from `--logfile` above as well.

src/plotman/graph.py
```diff
@@ -0,0 +1,263 @@
...
+        create_ax_plottime(ax, data_started_ended, window=3)
+
+        ax = plt.subplot(num_plots,1,4)
+        create_ax_plotcumulative(ax, data_started_ended)
+
+        ax.set_xlabel('Time (hours)')
+        f.savefig(figfile)
+
+
+if __name__ == '__main__':
```
altendky, line 248: Is it particularly needed for this to be directly runnable rather than just using the plotman CLI access?

altendky (commented):
It looks like there may have been some confusion with the catch up merge.  Take a look down through the diff and make sure all the changes are intended to be part of this PR.  Sorry for my role in this what with taking forever to catch up on the community contributions.

nikwl:
Hey @altendky, thanks for checking this out and for all of your feedback. I really only got a chance to bring the fork up to speed yesterday - I do mean to fix all the above issues but I'm a little pressed for time right now and I didn't get a chance to double check anything or actually test the code. When I do finally get everything working I'll "un-draft" the pr.

Thanks!

altendky:
Awesome, thanks.  Don't rush on my behalf.

nikwl:
Hey @altendky, sorry this has taken so long but I finally got around to updating this pr. If you could look it over when you get a chance that would be awesome. Thanks!

src/plotman/job.py
```diff
@@ -209,6 +204,8 @@ def get_running_jobs(
             # both identified as plot processes.  Only the child is
             # really plotting.  Filter out the parent.
 
+            pids = {process.pid for process in processes}
+            ppids = {process.ppid() for process in processes}
```
altendky, line 208: Are these (and above and below) needed?  Or perhaps they are just some confusion during some merges that can be removed?

src/plotman/plotman.py
```diff
@@ -95,8 +95,22 @@ def parse_args(self) -> typing.Any:
...
+        p_analyze.add_argument('--figfile', type=str, default=None,
+                help='figure to be created if logdir is passed')
```
altendky, line 103: I think that the changes to analyze are not related to the graphing addition anymore and can be removed?

altendky (dismissed):
I think we're close.  Still sorry for my slowness...  again.

altendky:
Oh, and add a changelog entry.

src/plotman/job.py
nikwl: Those must have been something I overlooked during the merges, there shouldn't have been any changes made to any files except plotman.py and graph.py. I dug a little deeper and it looks like several other files were changed, I'm not totally sure how, but I've updated them all to reflect the current state of the development branch, sorry about that.

altendky:
Maybe try https://github.com/nikwl/plotman/pull/1 to catch up and resolve the conflicts.

nikwl:
@altendky Thanks so much for the pr, that made things much, much simpler. I've updated the log parsing and tested it on my plots, seems to be working.

altendky:
Welp, a bit closer with the hints.  If you aren't sure how to handle these I can take a look later.

PR #626, opened 2021-05-22
add option to set tmp2 to dst

Ed-Yang:
The chia plot has its logic to choose the tmp2 directory, in some case if the dst is the local SSD, it is better to use dst SSD as tmp2 to avoid copy file from tmp2 to dst SSD.


Ed-Yang:
there is a pr #234 can achieve the same goal.

PR #627, opened 2021-05-22
Custom archive mode

altendky:
Further building on https://github.com/ericaltendorf/plotman/pull/88.

Draft for:
- [x] Continued cleanup
- [x] Maybe presets
- [x] Make `archive` subcommand work as well
- [x] Wiki the fact that there are configuration versions
  - https://github.com/ericaltendorf/plotman/wiki/Configuration#versions
- [x] Update wiki for new archival configuration
  - Being worked at https://gist.github.com/altendky/162700aac293ae72fdcb0e0fa4a33523
  - Copy over once released
- [x] The scripts presently take no command line parameters so the log lines for them are pretty useless
- [x] Logging of script output to catch errors
- [x] It is a bit silly to have to use quotes on the port like `"12000"`, even if understandable as an environment variable key in the YAML
- [x] Trailing `/`-agnosticness?
- [x] `set -evx` for disk space script ends up outputting via `stderr` in the interactive log
- [x] Tidy up commented out code etc at least


altendky:
Presumably we setup some recipes or maybe even some builtin presets but...

Example local config.

```yaml
        archive:
                index: 0
                env:
                    site_root: /farm/sites

                disk_space_script: |
                    #!/bin/bash
                    df -BK | grep " ${site_root}/" | awk '{ gsub(/K$/,"",$4); print $6 ":" $4*1024 }'

                transfer_script: |
                    #!/bin/bash
                    rsync --skip-compress plot --remove-source-files --inplace "${source}" "${destination}"
                transfer_process_name: rsync   
                transfer_process_argument_prefix: ${site_root}
```

Example rsyncd config.

```yaml
                index: 2
                env:
                    site_root: /farm/sites
                    url_root: rsync://altendky@server:12000/sites

                disk_space_script: |
                    #!/bin/bash
                    ssh altendky@server "df -BK | grep \" ${site_root}/\" | awk '{ gsub(/K\$/,\"\",\$4); print \$6 \":\" \$4*1024 }'"

                transfer_script: |
                    #!/bin/bash
                    relative_path=$(realpath --relative-to="${site_root}" "${destination}")
                    rsync --bwlimit=80000 --skip-compress plot --remove-source-files --inplace "${source}" "${url_root}/${relative_path}"
                transfer_process_name: rsync
                transfer_process_argument_prefix: ${url_root}
```

src/plotman/resources/plotman.yaml
```diff
@@ -59,31 +62,68 @@ directories:
...
+        # the `env` key.  plotman will additionally provide `source` and
+        # `destination` environment variables to the transfer script so it
+        # knows the specifically selected items to process.  plotman also needs
+        # to be able to generally detect if a transfer process is already
+        # running.  To be able to identify externally launched transfers, the
+        # process name and an argument prefix to match must be provided.  Note
+        # that variable substitution of environment variables including those
+        # specified in the env key can be used in both process name and process
+        # argument prefix elements but that they use the python substitution
+        # format.
```
rafaelsteil, line 86: I expanded a little bit the comments. PS: I think it's better to write it on the Wiki and add the link here. 

```yaml
# Archival configuration, used to move finished plots between
# directories or servers if you don't want to have them on the
# default local chia writes them. 
# 
# Whilst configuring archival is option, it's highly encouraged
# to do so, especially if you have more complex environments. It is 
# meant to distribute plots among multiple disks making the most
# usage of available space. 
# 
# If you don't plan to use archival, just comment this entire section. 
# 
# As of v0.4, archiving commands are highly configurable. The basic
# configuration consists of a script that checks available disk space
# and another for actually transferring the plots. Each can be specified
# as either a path to an existing script or inline script contents.
# 
# There's no technical difference between specifying an external
# script or adding it inline, it all comes down to your own preference. 
# 
# The default recipes provided by plotman should cover most common 
# scenarios, therefore you don't need to set anything else besides
# enabling it and choosing between local or remote and adjusting
# some of the default values. 
# 
# The `env` key it's a general purpose section to add any custom
# variable and value you want to make available to the underlying
# script. See the provided examples. 
# 
# On top of the keys and values that you set up in `env`, plotman provides 
# the variables `source` and `destination` to the transfer script 
# so it knows the specifically selected items to process.  
# 
# Also, plotman needs to be able to detect if a transfer process is already
# running. To be able to identify externally launched transfers, the
# process name and an argument prefix to match must be provided.  Note
# that variable substitution of environment variables including those
# specified in the env key can be used in both process name and process
# argument prefix elements but that they use the python substitution
# format.
```

src/plotman/resources/plotman.yaml
```diff
@@ -59,31 +62,68 @@ directories:
...
-                # the 4 machines, or 0, 1, 0, 1.
-                #   index: 0
+                # -- local rsync example setup
+
+                 env:
+                         command: rsync
+                         site_root: /farm/sites
+                 disk_space_script: |
+                         #!/bin/bash
+                         df -BK | grep " ${site_root}/" | awk '{ gsub(/K$/,"",$4); print $6 ":" $4*1024 }'
```
rafaelsteil, line 97: It would be useful to document the required output format.

rafaelsteil (commented):
Added a few comments. My main take is that it's powerful but really easy to mess it up. More error handling and proper logs would come super useful to help understand what's going on when things don't work as expected.

src/plotman/archive.py
```diff
@@ -87,40 +87,41 @@ def compute_priority(phase, gb_free, n_plots):
...
+        encoding='utf-8',
+        env={**os.environ, **arch_cfg.environment()},
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+
+    for line in completed_process.stdout.splitlines():
+        split = line.split(':')
+        if len(split) != 2:
+            # TODO: warning?  or something...
```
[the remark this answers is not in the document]
mjsr: Would it make sense to test all the archive configurations during startup of one small hidden file (e.g., `.plotman-archive-test`)?

This could be triggered before the main plots command or the archive command at least to ensure archiving will work and can help folks quickly test their config. Alternatively, a dedicated archive-test command would print more verbose errors regarding the archiving configuration.

src/plotman/plotman.py
```diff
@@ -137,120 +137,121 @@ def main():
...
+                print('...starting archive loop')
+                firstit = True
+                while True:
+                    if not firstit:
+                        print('Sleeping 60s until next iteration...')
+                        time.sleep(60)
+                        jobs = Job.get_running_jobs(cfg.directories.log)
+                    firstit = False
+
+                    archiving_status, log_messages = archive.spawn_archive_process(cfg.directories, jobs)
```
rafaelsteil, line 193: Missing `arch_cfg`

src/plotman/configuration.py
```diff
@@ -82,6 +210,9 @@ def get_dst_directories(self):
 
         return self.dst
 
+    def create_log_path(self, group, time=None):
```
rafaelsteil, line 213: Looks like `time` shouldn't have `None` as default, as it's required by the method

src/plotman/archive.py
```diff
@@ -9,37 +9,65 @@
...
+                    f'Archiving log file already exists, skipping attempt to start a'
+                    f' new plot: {log_file_path!r}'
+                )
+                return (False, log_messages)
+            except FileNotFoundError as e:
+                message = (
+                    f'Unable to open log file.  Verify that the directory exists'
+                    f' and has proper write permissions: {log_file_path!r}'
+                )
+                raise Exception(message) from e
```
rafaelsteil, line 55: It would be prudent to check it on startup rather than here. I added a snipped in another comment.

src/plotman/plotman.py
```diff
@@ -137,120 +137,121 @@ def main():
...
+            # Directories report
+            elif args.cmd == 'dirs':
+                print(reporting.dirs_report(jobs, cfg.directories, cfg.scheduling, get_term_width()))
+
+            elif args.cmd == 'interactive':
+                interactive.run_interactive(cfg=cfg)
+
+            # Start running archival
+            elif args.cmd == 'archive':
+                print('...starting archive loop')
```
rafaelsteil, line 184: Add a check to make sure we'll be able to actually write the logs before even starting: (related to another comment I made)

```python
if not path.exists(cfg.directories.log) or not os.access(cfg.directories.log, os.W_OK):
    print(f"Error: Directory '{cfg.directories.log}' doesn't exist or isn't writable")
    sys.exit(1)
```

src/plotman/archive.py
```diff
@@ -9,37 +9,65 @@
...
+
+            log_file_path = dir_cfg.create_log_path(group='transfer', time=pendulum.now())
+
+            # TODO: CAMPid 09840103109429840981397487498131
+            try:
+                open_log_file = open(log_file_path, 'x')
+            except FileExistsError:
+                log_messages.append(
+                    f'Archiving log file already exists, skipping attempt to start a'
+                    f' new plot: {log_file_path!r}'
```
rafaelsteil, line 47: Considering the context, I think you meant `skipping attempt to start a new archive` than `plot`

src/plotman/archive.py
```diff
@@ -9,37 +9,65 @@
...
+            with open_log_file:
+                # start_new_sessions to make the job independent of this controlling tty.
+                p = subprocess.Popen(**args,
                     shell=True,
-                    stdout=subprocess.DEVNULL,
+                    stdout=open_log_file,
                     stderr=subprocess.STDOUT,
                     start_new_session=True)
-            log_message = 'Starting archive: ' + cmd
+            log_messages.append('Starting archive: ' + args['args'])
```
rafaelsteil, line 70: Include the log path: 

```python
log_messages.append('Starting archive: ' + args['args'] + f". Saving logs to {log_file_path}")
```

rafaelsteil (commented):
1. It doesn't seem that `src/plotman/resources/target_definitions.yaml` is being copied to the config dir when running `plotman config generate`. I'd expect it to be the case, since users will potentially need / want to set custom values

2. Running `plotman archive` from a fresh install crashes with the following exception, as `arch_cfg` is not being passed:

```
...starting archive loop
  File "/usr/local/bin/plotman", line 33, in <module>
  File "/Users/rafaelsteil/Desktop/custom-archive/plotman/src/plotman/plotman.py", line 193, in main
    archiving_status, log_messages = archive.spawn_archive_process(cfg.directories, jobs)
TypeError: spawn_archive_process() missing 1 required positional argument: 'all_jobs'
```

src/plotman/archive.py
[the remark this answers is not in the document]
altendky: I had considered a `plotman config check` command or such as a separate activity.  Doing an automatic precheck on startup for some things may also be worthwhile, perhaps, but I think I would leave that for a separate effort.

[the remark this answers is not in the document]
altendky: (where `check` would involve doing stuff like writing a fake log (and removing it) and so on like you suggested for archiving, not just static checks of the config)

src/plotman/plotman.py
altendky:
```suggestion
                    archiving_status, log_messages = archive.spawn_archive_process(cfg.directories, cfg.archiving, jobs)
```

src/plotman/archive.py
altendky:
```suggestion
                    f'Archiving log file already exists, skipping attempt to start a'
                    f' new archive transfer: {log_file_path!r}'
```

src/plotman/archive.py
altendky: I think this matches the plot starting.  But yeah, this is a great way to make the log line more useful, thanks.

```suggestion
            log_messages.append(f'Starting archive: {args["args"]} ; logging to {log_file_path}')
```

src/plotman/configuration.py
altendky: Uh, yeah...  changed my mind on this and didn't remove the default.

```suggestion
    def create_log_path(self, group, time):
```

altendky (commented):
A lot of good catches here, thanks.  Here are responses to some but not all of them.

`target_definitions.yaml` is specifically for the builtin presets.  Users can write their own target definitions in the regular plotman configuration at `archiving: target_definitions: my_target:` (or whatever name they choose).  So, I'll lump this into the 'needs documentation' task, unless I have missed your point.  Probably a commented out example in the `plotman.yaml` as well, I guess?

PR #630, opened 2021-05-22
Add a config option to force using the current dst directory as the t…

bkmaibach:
…mp2 directory

### DIFF
--- a/src/plotman/manager.py
+++ b/src/plotman/manager.py
@@ -128,6 +128,9 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
             if dir_cfg.tmp2 is not None:
                 plot_args.append('-2')
                 plot_args.append(dir_cfg.tmp2)
+            elif plotting_cfg.use_dst_as_tmp2:
+                plot_args.append('-2')
+                plot_args.append(dir_cfg.dst)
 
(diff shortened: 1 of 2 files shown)


src/plotman/manager.py
```diff
@@ -128,6 +128,9 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
             if dir_cfg.tmp2 is not None:
                 plot_args.append('-2')
                 plot_args.append(dir_cfg.tmp2)
+            elif plotting_cfg.use_dst_as_tmp2:
+                plot_args.append('-2')
+                plot_args.append(dir_cfg.dst)
```
xsilmarx, line 134: When tmp2 is same as dst dir, chia plotter won't copy the final file, it will just rename it.
Bases on this it would be better to use the value of 'dstdir' variable because if you have multiple dst dirs, these way you make sure the tmp2 is the same as dstdir for this specific plot, and thus avoid an additional copy at the end.

src/plotman/manager.py
```diff
@@ -128,6 +128,9 @@ def maybe_start_new_plot(dir_cfg, sched_cfg, plotting_cfg):
...
```
altendky, line 133: Isn't `.dst` a list not a single directory?

altendky:
Closing since the originating repository was deleted.  This could be reopened and comments responded to if still interested.  Side note, please target PRs at the `development` branch.

PR #632, opened 2021-05-22
it would be nice to have a copy or move function as an archive option. 

lnb2g:
rsync is nice and would do the job, but if you plot a lot, it become a bottleneck. Please add a cp/mv between 2nd temp drive and dst. Thanks


altendky:
Was this supposed to be an issue and not a pull request?  Also, there has been a lot of rework on archiving.

altendky:
`tmp`, `2`, and `dst` are all handled by the chia plotting process.  They do not use rsync.  plotman archiving is being reworked in https://github.com/ericaltendorf/plotman/pull/627 if you want to look.  The OP has a couple links to in-work documentation.

PR #642, opened 2021-05-23
Archive balancing based on rsync temp files

randellhodges:
Proof of concept. Makes a few assumptions that might not be valid across all the different platforms and won't work with some of the other changes here on opening up the archive process to other commands.

Idea:

Say you have 10Gbe (or more) and you have multiple plotman instances running on your network and you want to archive to a single server with more disks shared than plotman instances. When the archive routine runs, it'll get the list of destinations that has free disk space (as it did before). It'll then look to see if rsync temp file pattern is there, indicating a different plotman instance is archiving there. If it finds it, it'll try a different destination. This will balance the copies across disks.

It still has 1 archive process per plotman instance so it won't help in those situations.

Edit: Resuming partial transfers would probably fail because it might not go back to the same drive as before, so that invalidates the -P option on rsync. A way around that would be to adjust the logic to look for a file that has the same base name as the copy source. I might look into that.

NOTE: I have made these changes against the development branch, but I don't run the development branch at home.  For the moment, I'm just hand patching changes into the main branch and when they work, I put them here. If you try to merge these or run this fork directly, there might be an issue. If there is, let me know and I'll fix it.

### DIFF
--- a/src/plotman/archive.py
+++ b/src/plotman/archive.py
@@ -163,22 +163,41 @@ def archive(dir_cfg, all_jobs):
                  space > 1.2 * plot_util.get_k32_plotsize()]
 
     if len(available) > 0:
+        available = sorted(available)
+        candidates = []
+
         for candidate in available:
+            # The last time thru the loop we found a resumable temp file so
+            # break ths loop now
+            if archdir:
+                break
+
+            # Observed pattern is:
+            #   plot-k32-something.plot => .plot-k32-something.plot.XXXX
             # Check if there is an rsync temp file here. TODO maybe use find or a regex
             # pattern or make the pattern configurable, or even find better documentation on rsync temp files.
             # Making it configurable would fit better with upcoming changes to opening up the archive process,
             # Assuming that custom process created a tmp file and cleaned up after itself
-            temp_cmd = ('ssh %s@%s ls -h %s/.plot*.plot.* 2>/dev/null' %
+            temp_cmd = ('ssh %s@%s ls -1 %s/.plot*.plot.* 2>/dev/null' %
                             (dir_cfg.archive.rsyncd_user, dir_cfg.archive.rsyncd_host, candidate[0]) )
 
             with subprocess.Popen(temp_cmd, shell=True, stdout=subprocess.PIPE) as proc:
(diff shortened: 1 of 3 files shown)


florianweyandt1337:
Thanks for this. I'm already seeing contention on one drive even with 1Gbe and 2 Plotting machines. 
May I suggest another improvement to better balance disk usage?
Changing
`(archdir, freespace) = candidates[0]`
to
`(archdir, freespace) = candidates[random.randint(0, len(candidates))]`
I find it annoying to have one disk filled and the others not and I like a more balanced graph of disk usage :)

randellhodges:
I'll look at adding an option for this, but I don't plan on spending too much effort on this because from what I'm reading, it will be invalidated once the archiving rewrite is implemented. Maybe some things here will spur ideas during the rewrite.

I think my title is wrong. As it stands, it isn't balancing, so much as distributing based on a disk in use :)

randellhodges:
Another thing it might need, an option to allow transfers to a destination where there are already active transfers in progress? Not sure what is worse.

Say you have 3 machines running plotman, and all 3 have finished a plot and they all start an archive to 3 different mounts. With your 10Gbe connection, reading from an SSD, you are at the mercy of the HDD at the destination. Let's say you get a solid 100MB/s write speed. You are looking at 15 minutes or so to write out that file.

If you only had 3 destinations and they are all in "use", the current version of this code will block and wait if another plot finishes. I feel this is better because it helps prevent fragmentation at the destination, allows the other transfer to finish quicker. But I just wanted to point that out.

altendky:
I didn't come up with this, but does `--preallocate` on rsync not cut it?  I imagine, at least, that it would cut down the race condition to being a pretty short time and that on the next loop (20s by default in interactive) a new transfer would be launched to a drive that now looks to have enough space.

altendky:
My apologies for being slow in processing contributions.  I am trying to catch up and focus in on the items that need work.  It seems to me that the `--preallocate` option, which is now enabled by default, may fill the high level need here.  If that is not the case then let's discuss the details.  We can reopen this if we decide it is still a good improvement.  Thanks for your effort to improve the project.

PR #643, opened 2021-05-23
Use tmp as dst when it is also a dst

djdjukic:
Modified manager.py to detect when a selected tmpdir is also configured as a dst, and, in that case, use it as the dstdir in order to avoid unneccessary copying. Useful for plotting directly to HDD.


syadnom:
upvote!

altendky:
So for some tmps you want separate dsts and for other tmps you want dst-on-tmp?

djdjukic:
@altendky Yes. I want to plot in two ways:

1. conventional SSD-to-HDD plotting - SSD tmp, HDD dst
2. whole plotting process entirely on HDD - HDD tmp, same HDD dst

What we have now is random tmp, random (HDD) dst.

A good addition to this PR would be a way to set tmpdir_max_jobs individually per tmp, to allow us to set 1 max plot for HDD tmps and 2 or more for SSD tmps.

altendky:
You already have the tmp overrides to cover the extra bit you mentioned of per-tmp plot limits.  I'll think about this but my general plan is to long term work towards being able to specify multiple workflows.  In your case you would have two.

djdjukic:
Multiple workflows - that sounds even better. Though I think this commit would still make a good default for mixed setups. Thanks for the tip about tmp overrides, I glossed over that setting, it does just what I want.

altendky:
@djdjukic, I took the liberty of doing a catch up merge for you.  Could you try it out and make sure I got it right?  If so, we can go ahead and merge this.  Thanks for your contribution and my apologies for being so slow about this.

djdjukic:
@altendky I have now tested the updated branch, and this feature indeed works as it should. Thank you!

PR #704, opened 2021-05-30
Include copy time on plotman analyze

ReaktorDTR:
Added: Copy time column in analyzer
Updated: Total time includes Copy time

### DIFF
--- a/src/plotman/analyzer.py
+++ b/src/plotman/analyzer.py
@@ -1,7 +1,6 @@
 import os
 import re
 import statistics
-import sys
 
 import texttable as tt
 
@@ -18,6 +17,7 @@ def analyze(logfilenames, clipterminals, bytmp, bybitfield):
             n_sorts = 0
             n_uniform = 0
             is_first_last = False
+            total_time = 0
 
             # Read the logfile, triggering various behaviors on various
             # regex matches.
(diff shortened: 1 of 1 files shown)


src/plotman/analyzer.py
```diff
@@ -89,14 +90,27 @@ def analyze(logfilenames, clipterminals, bytmp, bybitfield):
...
+
+                # Copy Time.  Sample log line:
+                # Copy time = 958.194 seconds. CPU (6.290%) Sat May 22 17:39:57 2021
+                m = re.search(r'^Copy time = (\d+.\d+) seconds.*', line)
+                if m:
+                    copy_time = float(m.group(1))
+
+                # Rename plot.  Sample log line:
+                # Renamed final file from "/mnt/d1/a4.plot.2.tmp" to "/mnt/d2/a4.plot"
+                m = re.search(r'^Renamed final file.*', line)
```
altendky, line 103: Is this present regardless of what combinations of tmp/tmp2/dst are specified?

ReaktorDTR: Yes, and if tmp same as dst it will show 0 sec because of simple renaming.

altendky:
I'm a bit hesitant to have total time in analyze mean a different thing than total time in the logs.  Any ideas what else would be a good name for the combination?  Having total, copy, and combined time would allow the output code to stay more like before but just add the two new bits.

Do you have a log you could share that shows the copy time?  My most recent logs don't include it.  They should have been from 1.1.7.  I have tmp the same as dst so there's no copying so it's understandable.

ReaktorDTR:
It might be called Summary. I don't have logs for 1.1.7. When I got the first numbers in the Copy Time column I made a decision to switch from separate folders to one for all 😄

altendky:
How about we just go literal with `start + copy`.  I think that still ends up about the same length as the data in the table cells.  If there's inspiration later it can change.

PR #707, opened 2021-05-30
feat: reread config file on each pass of plot loop

wtfrank:
Allows settings such as number of threads to be tweaked without restarting plotman.

### DIFF
--- a/src/plotman/plotman.py
+++ b/src/plotman/plotman.py
@@ -142,6 +142,8 @@ def main():
     if args.cmd == 'plot':
         print('...starting plot loop')
         while True:
+            config_text = configuration.read_configuration_text(config_path)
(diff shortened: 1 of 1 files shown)


PuppyLover101 (approved):

altendky:
This would at least need to be configurable.  I share my configs between multiple machines and use git to transfer between and so various changes could get loaded that do not make sense while I am processing through that.  Also, what should happen if the reload fails?

mk01:
@altendky 
of course old config should be retained.

```
            try:
                newcfg = configuration.get_validated_configs(configuration.read_configuration_text(config_path), config_path)
                cfg = newcfg
            except Exception as ex:
                print('... while reloading config: %s' % (str(ex)))
```

PR #739, opened 2021-06-05
Show ploting in the gui

Chiazz:
Hi pls that possible to add a view of ploting in the gui like manual launching?


altendky:
I'm not sure what you are referring to, sorry.  Also, perhaps you want to open [a new discussion](https://github.com/ericaltendorf/plotman/discussions/new) instead of a pull request (which is for when you want to submit a code change).

PR #751, opened 2021-06-07
fix: is_plotting_cmdline not recognized chia process

Indexyz:
In NixOS, the chia plotting process be like 

```
/nix/store/q6gfck5czr67090pwm53xrdyhpg6bx67-python3-3.8.9/bin/python3.8 /nix/store/90s76gnjw3vahj3yhilwq8jfmg4d3bcw-chia-1.1.5/bin/.chia-wrapped plots create -k 32 -b 4000 -t /mnt/cache -d /mnt/plots -r 4 -u 128 -f fpk -p ppk
```

So the cmdline is not ending with `chia`


altendky:
This needs to target the `development` branch.  We should have noted this in a PR template to remind people (including myself) but I just now got around to [starting to add that](https://github.com/ericaltendorf/plotman/pull/753).  Anyways...  the code isn't actually different so the change is still relevant.
 https://github.com/ericaltendorf/plotman/blob/e2595074491622618526fd9627adf74d2c7fad5f/src/plotman/job.py#L31-L39

I hesitate a bit to make it as broad as `'chia' in` since then any `'chia'` anywhere in the path then qualifies it.  Though as I look at it, the same could be said for the `'python'` check as well.  `:|`  So I guess this would be fine as is.  I might make it check only the file name instead of the full path in the future (for both the `'python'` and `'chia'` checks).

PR #758, opened 2021-06-08
Tmp dir overrides

racergoodwin:
This request adds more options to the tempdir overrides, allowing for different start criteria of multiple tempdirs.


altendky:
Could you do a catch up merge and I'll take another look?  There are type hints now so if you have to wait for me to approve again after the update, you can give `tox -e check-hints-py39` (or whatever python version you have) a try locally for feedback on that.

One point I'm not sure on, like not sure which I actually prefer, is whether the override should be allowed to be incompletely specified.  I understand reducing repetition, but does it really make sense to be able to override just the phase major, for example?

racergoodwin:
Sure thing.
Yeah, it could cause confusion if it isn't necessary to specify major and minor phases. Will change that behaviour.

altendky:
I also started adding brief lines in the changelog with links to the PRs.  Could you please add one for this while you're at it?

src/plotman/manager.py
```diff
@@ -53,25 +53,39 @@ def phases_permit_new_job(phases: typing.List[job.Phase], d: str, sched_cfg: plo
...
+            stagger_phase_limit = curr_overrides.tmpdir_stagger_phase_limit
+        # Check for and assign stagger phase limit override
         if curr_overrides.tmpdir_max_jobs is not None:
             max_plots = curr_overrides.tmpdir_max_jobs
-    if len(phases) >= max_plots:
+        
+    milestone = job.Phase(major,minor,)
+    
+    # Check if phases pass the criteria
+    if len([p for p in phases if p < milestone]) >= stagger_phase_limit or len(phases) >= max_plots:
```
altendky, line 85: I don't have a formal argument for this, but I think I would rather keep this more in the form it was before.

```suggestion
    if len([p for p in phases if p < milestone]) >= stagger_phase_limit
        return False

    if len(phases) >= max_plots:
```

src/plotman/manager.py
```diff
@@ -53,25 +53,39 @@ def phases_permit_new_job(phases: typing.List[job.Phase], d: str, sched_cfg: plo
...
         return False
-
-    return True
+    else:
+        return True
```
altendky, line 88:
```suggestion

    return True
```

src/plotman/manager.py
```diff
@@ -53,25 +53,39 @@ def phases_permit_new_job(phases: typing.List[job.Phase], d: str, sched_cfg: plo
...
+            minor = curr_overrides.tmpdir_stagger_phase_minor
+        # Check for and assign stagger phase limit override
+        if curr_overrides.tmpdir_stagger_phase_limit is not None:
+            stagger_phase_limit = curr_overrides.tmpdir_stagger_phase_limit
+        # Check for and assign stagger phase limit override
         if curr_overrides.tmpdir_max_jobs is not None:
             max_plots = curr_overrides.tmpdir_max_jobs
-    if len(phases) >= max_plots:
+        
+    milestone = job.Phase(major,minor,)
```
altendky, line 82:
```suggestion
    milestone = job.Phase(major,minor)
```

src/plotman/manager.py
```diff
@@ -81,9 +95,9 @@ def maybe_start_new_plot(dir_cfg: plotman.configuration.Directories, sched_cfg:
     youngest_job_age = min(jobs, key=job.Job.get_time_wall).get_time_wall() if jobs else MAX_AGE
     global_stagger = int(sched_cfg.global_stagger_m * MIN)
     if (youngest_job_age < global_stagger):
-        wait_reason = 'stagger (%ds/%ds)' % (youngest_job_age, global_stagger)
+        wait_reason = 'global stagger (%ds/%ds)' % (youngest_job_age, global_stagger)
     elif len(jobs) >= sched_cfg.global_max_jobs:
-        wait_reason = 'max jobs (%d) - (%ds/%ds)' % (sched_cfg.global_max_jobs, youngest_job_age, global_stagger)
+        wait_reason = 'global max jobs reached (%d)' % (sched_cfg.global_max_jobs)
```
altendky, line 100: The stagger seconds was added not too long ago.  I personally plan on going the other way and considering how to indicate the status of all restrictions at all times.  Was there a particular need to remove this?

```suggestion
        wait_reason = 'global max jobs reached (%d) - (%ds/%ds)' % (sched_cfg.global_max_jobs, youngest_job_age, global_stagger)
```

Side note, and no need to change it now, but this PR probably should have been at least four.  Offhand, the new per-tmp phase stagger, the addition of global, removal of the time-stagger indication here, and moving of the overrides to scheduling.  Just a thought for next time.  Small PRs are quicker to verify and merge (and undo if needed).

src/plotman/manager.py
```diff
@@ -92,7 +106,7 @@ def maybe_start_new_plot(dir_cfg: plotman.configuration.Directories, sched_cfg:
                 for (d, phases) in eligible ]
 
         if not eligible:
-            wait_reason = 'no eligible tempdirs (%ds/%ds)' % (youngest_job_age, global_stagger)
+            wait_reason = 'waiting for phase match'
```
altendky, line 109: Is this actually limited to indicating phase-related restrictions?  I quickly checked `phases_permit_new_job()` and it looks like it also checks the process limit.

https://github.com/ericaltendorf/plotman/blob/d52f51600736dbb2a1635ab80eede606949f1903/src/plotman/manager.py#L47-L74

src/plotman/manager.py
```diff
@@ -81,9 +95,9 @@ def maybe_start_new_plot(dir_cfg: plotman.configuration.Directories, sched_cfg:
     youngest_job_age = min(jobs, key=job.Job.get_time_wall).get_time_wall() if jobs else MAX_AGE
     global_stagger = int(sched_cfg.global_stagger_m * MIN)
     if (youngest_job_age < global_stagger):
-        wait_reason = 'stagger (%ds/%ds)' % (youngest_job_age, global_stagger)
+        wait_reason = 'global stagger (%ds/%ds)' % (youngest_job_age, global_stagger)
```
altendky, line 98: If there's something to clarify wouldn't it be that this is for the time-stagger not the phase-stagger?  Though the reporting of seconds adjacent to the `stagger` comment it seems to lead that way.

altendky (changes requested):
Since this is a backwards incompatible configuration change we should bump `plotman.configuration.expected_major_version` to `2`.  We'll need to add some commentary to the [wiki configuration page](https://github.com/ericaltendorf/plotman/wiki/Configuration#versions) as well.

src/plotman/manager.py
racergoodwin: No worries. I will shift it back.
To my eye, it looked neater and was easier to understand with less repetition

src/plotman/manager.py
racergoodwin: It seems that the only relevant need for the seconds to be displayed is for the global stagger m setting.
To me, it looks poor, having a clock that goes passed it's target and keeps ticking when it is no longer relevant.
Maybe keep showing youngest_job_age and remove global_stagger, once global_stagger has passed?
I had thought about adding an option of putting individual running time counters (hh:mm) in front of each running plot so people would know their overall duration, but as a different PR.

racergoodwin:
Thanks for looking at this. I will get onto them all this evening.

src/plotman/manager.py
altendky: It lets you know how far past the time stagger you have gone based on some other restriction.  Going beyond the time stagger isn't inherently bad. 
 I specifically set my time stagger a bit less than even spacing would be based on my total jobs and job duration.  Anyways, to satisfy "everyone" I expect I'll need to make most of the pieces optional.  There's a PR adding a couple lines and another adding a couple columns and...  `:]`  And I don't super want to do that in curses when I've already started a switch to prompt_toolkit.

altendky:
Just a quick note, the failing macOS build looked like a general CI setup failure of some sort so I reran all the jobs.  Looks like they pass now.

altendky:
I unresolved a couple "conversations" that were marked resolved without related code changes or replies.  Neither seem to be related to the core of this PR so it would be good to not hold the whole thing up for too long on them.  `:]`

racergoodwin:
Fair enough. I thought I'd clicked on all the commit suggestions?

src/plotman/manager.py
racergoodwin: all wait_reason defines returned to original

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `plotman kill` doesn't leave any temporary files behind anymore.
   ([#801](https://github.com/ericaltendorf/plotman/pull/801))
 ### Added
+- optional tmpdir_overrides added for each specified tmpdir
+  ([#758](https://github.com/ericaltendorf/plotman/pull/758))
```
altendky, line 14: The "overrides" were already there, right?  The phase limit control was added.

```suggestion
- tmp directory overrides moved to `scheduling:` `tmp_overrides:`.
  ([#758](https://github.com/ericaltendorf/plotman/pull/758))
- Per tmp directory phase limit control added to `scheduling:` `tmp_overrides:`.
  ([#758](https://github.com/ericaltendorf/plotman/pull/758))
```

altendky (commented):
Thanks again for the work here.  I think what's here is good and I can handle the somewhat separate layers of configuration structure changes that we talked about.  Would you be comfortable with me adding that into this PR?

PR #783, opened 2021-06-13
Opinionated docker build

graemes:
This is a 'opinionated' docker build.  It:
- utilises a multi-stage build to minimise space & potential security issues ;
- uses an internal user/group id which should be mapped to a user/group on the host (all bound directories should be owned by this user.  Again this is for process isolation for security ;
- runs 'plotman plot' on startup ;


Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+&& chmod +x install.sh 
+
+WORKDIR /chia-blockchain
+# Placeholder for patches
+RUN /usr/bin/sh ./install.sh
+
+COPY . /plotman
+
+RUN . ./activate \
+&& pip install -e /plotman \
```
altendky, line 21: If you don't do `-e` (`--editable`) here then you don't have to copy over `/plotman` into the final stage.  It will just be part of the chia environment.  There are a few other references that I think would go away as well.

build-docker-plotman.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
altendky, line 1: Do we really want to think about not-bash?

```suggestion
#!/bin/bash
```

build-docker-plotman.sh
```diff
@@ -0,0 +1,20 @@
...
+
+DOCKER_REGISTRY="graemes"
+LOCAL_REGISTRY="registry.graemes.com/graemes"
```
altendky, line 4: We probably don't want to reference your personal stuff here?  At least not unless we are going to host it there for everyone else to use?

altendky (changes requested):
I need to go over this in more detail in a few different areas, but here's a start to the conversation.  Thanks for the work here and my apologies for the delay in getting to it.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+&& cd chia-blockchain \
+&& git submodule update --init mozilla-ca
+
+WORKDIR /chia-blockchain
+# Placeholder for patches
+RUN /bin/bash ./install.sh
+
+COPY . /plotman
+
+RUN ["/bin/bash", "-c", "source ./activate && pip install /plotman && deactivate"]
```
altendky, line 20: Doesn't really matter, but activation won't outlast the shell so deactivation doesn't do much here.

altendky (commented):
Could you write up a 'how to use this' in the wiki?  Or just start with a less "formal" comment here or whatever.  I'm thinking there's some setup for the volumes?

altendky:
Oops, and a changelog entry.

PR #821, opened 2021-07-02
introduce black

altendky:
Draft for:
- [x] Actually formatting the code (soon...)


altendky:
The message below is intended to be placed in each open PR prior to merging to notify the contributors about this change and to offer to help deal with it.

---

"Soon" we will be using [Black](https://black.readthedocs.io/en/stable/?badge=stable) to format the code (https://github.com/ericaltendorf/plotman/pull/821).  This isn't because I'm a super big fan of the actual formatting but rather because it is a well accepted formatter that gets mostly tidy code and avoids discussion and thinking about how to format stuff.

I understand that this can be a hassle for outstanding branches.  I apologize for this hassle and am happy to let you handle it yourself, to help you through handling it, or just to update your branch for you.  Just let me know if you want me to do something about it.

PR #832, opened 2021-07-06
- fix Job.init_from_logfile for MadMax, currently taking 3s / process

mk01:


altendky:
Thanks for the contributions.  FYI, I've started refactoring this to have completely separate modules for support of each plotter.  There are more coming.

PR #833, opened 2021-07-06
- fix Job.is_plotting_cmdline for MadMax

mk01:
- only basename of the process name should be compared
    (it can contain full path to binary, or './' etc)


altendky:
Thanks here as well, I'll get these merged into the release branch.

Issue #121, opened 2021-04-17
Add option to stop creating new plots as dst becomes full


mikehw:
Currently free space on destination directories is not evaluated when determining which destination directory to select.

Having the option to not plot to full drives would be nice for users that aren't utilizing archiving and want to stop plotting once the destination directories are full. This should be off by default since archiving should free up space while the plot is being created.

13thProgression:
When using the development branch with multiple destination drives, plotman stopped scheduling plots for a full drive. I had archive disabled. Have you tested the development branch for this capability?

BasilHorowt:
> When using the development branch with multiple destination drives, plotman stopped scheduling plots for a full drive.

Hmm I'm not sure this is actually true... I wonder if something else stopped it from plotting? @altendky is there actually anything in dev branch that would do this?  Assuming it is still an open issue, @13thProgression you may want to subscribe to https://github.com/ericaltendorf/plotman/issues/176

As for the original issue

>Currently free space on destination directories is not evaluated when determining which destination directory to select.

sounds different from 

>want to stop plotting once the destination directories are full

The former could be seen as a request to factor in DST empty space when choosing which drive to plot to, while the later is the same issue as https://github.com/ericaltendorf/plotman/issues/176

@mikehw care to chime in with more details?

PR #842, opened 2021-07-09
Add updated configuration to exclude full dst directories

roxleopardo:
I updated the existing pull request #121 for working with the current plotman configuration and features


altendky:
What situation does the newer (in v0.4) simple local rsync config not address?

roxleopardo:
> What situation does the newer (in v0.4) simple local rsync config not address?

Sorry, I never used archiving configuration. Using local rsync solved the problem. Thank you!

PR #878, opened 2021-07-18
Standardise logging for interactive and plot/archive 'daemons'

graemes:
This creates the same log entries for plotting and archive in the plotman root log.  It also move the disk_space script output to a separate log file to reduce the noise in the root log.  Eventually would like to modify the interactive UI to read plot/archive updates from the log file to provide visibility of interactive and daemon invocations.


altendky:
Thanks as always, as I try to catch up a bit here...

PR #898, opened 2021-08-06
Docker improvements

Artanicus:
Use supervisord to run both plotting & archiving by default. This can easily be overridden by volume mounting a replacement `/srv/supervisord.conf`

- Suggest default tagging the image as `latest` since it's the Docker default and often has a special meaning in orchestration.
- Update Chia reference to latest stable
- Avoid clobbering $UID which can have unintended consequences

### DIFF
--- a/build-docker-plotman.sh
+++ b/build-docker-plotman.sh
@@ -7,15 +7,15 @@ BASE_CONTAINER="ubuntu:20.04"
 CHIA_GIT_REFERENCE="1.2.3"
 
 # The UID/GID should match the 'chia' owner of the directories on the host system
-bUID=10001
-bGID=10001
+build_UID=10001
+build_GID=10001
 
 docker build . \
 	--squash \
 	--build-arg BASE_CONTAINER=${BASE_CONTAINER} \
 	--build-arg CHIA_GIT_REFERENCE=${CHIA_GIT_REFERENCE} \
-	--build-arg UID=${bUID} \
-	--build-arg GID=${bGID} \
+	--build-arg UID=${build_UID} \
+	--build-arg GID=${build_GID} \
 	-f docker/Dockerfile \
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN /bin/bash ./install.sh
 
 COPY . /plotman
 
-RUN ["/bin/bash", "-c", "source ./activate && pip install /plotman && deactivate"] 
+RUN ["/bin/bash", "-c", "source ./activate && pip install /plotman && pip install supervisor && deactivate"]
```
altendky, line 20: We are already installing into an existing env which pip isn't the best at.  But we can still let pip know all the additional things it needs to install at once.

```suggestion
RUN ["/bin/bash", "-c", "source ./activate && pip install /plotman supervisor && deactivate"]
```

build-docker-plotman.sh
```diff
@@ -2,22 +2,20 @@
...
 BASE_CONTAINER="ubuntu:20.04"
-CHIA_GIT_REFERENCE="1.1.7"
+CHIA_GIT_REFERENCE="1.2.3"
 
 # The UID/GID should match the 'chia' owner of the directories on the host system
-UID=10001
-GID=10001
-
-docker rmi ${LOCAL_REGISTRY}/${PROJECT}:${TAG}
+bUID=10001
```
altendky, line 10: Not clobbering `UID` sounds good.  Does the `b` refer to anything in particular?

altendky (commented):
Just a couple details here from me, for now.  @graemes, want to chime in here?

build-docker-plotman.sh
Artanicus: Just build_UID essentially. I'll make it more descriptive :-)

graemes:
Looks good to me.

altendky:
Thanks for this and my apologies for the extended delay in handling it.

PR #913, opened 2021-08-28
Fixup tmp cleanup

altendky:
https://github.com/ericaltendorf/plotman/discussions/912

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 ### Fixed
+- Temp files are correctly identified for cleanup.
+  ([#912](https://github.com/ericaltendorf/plotman/pull/913))
 ### Added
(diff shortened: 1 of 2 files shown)


altendky:
I'm getting file names such as `plot-k32-2021-08-28-15-05-0b9e4436b8279739d582e3213277040bc9bea4c40bc72aff552fcda56f775d46.p1.t1.sort_bucket_9.tmp` from madMAx anyways.  The existing pattern doesn't match this.

altendky:
And `plot-k25-2021-08-28-15-19-2762fbdc8cc76b1e90d646db6e199522fed8ef5db8602b253e2b1fd1854a7263.plot.p1.t1.sort_bucket_079.tmp` from chiapos.

altendky:
err, need a changelog entry

ericaltendorf (approved):
Looks good.  FWIW I'm on board with creating a new directory for each plot job and using that for isolation, then we could use full wildcard.

altendky:
Yep, that's somewhere on the list...

Issue #827, opened 2021-07-03
Option to donate to the authors


ruXlab:
Thanks for the hard work and maintenance of the project

Why don't you create a wallet for donations. It will be a way to say thanks back to the development team.

altendky:
We can see what @ericaltendorf says, but I personally have chosen in my open source work to not do that.  Since we have many contributors there would be some effort involved in finding a mutually satisfactory distribution of any donations.  To be clear, I fully support those open source developers that do get income from their work.  I just personally have chosen not to for various reasons.

In the past when people have offered me money I have suggested they pick a charity to donate to instead.  In one case they actually did.  So, my suggested action for this ticket would be to document this in some form in the readme.  But, we can see who else chimes in.

Also, you're welcome.  I'm glad it helps you out.  `:]`

altendky:
I started working for Chia Network this past Monday.  I have not been tasked with working on plotman for the company but none the less I figured it was appropriate to share this given the above comments about donations.  I appreciate the community for all the help and opportunities they've shared.  Let's see how great of a tool we can all build here.  `:]`

PR #916, opened 2021-08-30
BladeBit

altendky:


altendky:
This has been run only briefly and on a system with only 64GB of RAM so a single job hasn't even been run to completion.  But, it was able to launch and identify the process.  Remember that BladeBit won't use a tmp directory and so tmp directory phase limits and such won't apply to BladeBit.  Rather, you would generally set the global max jobs to 1.  A global phase limit has been discussed but has not been implemented.

jonathanhatch:
> This has been run only briefly and on a system with only 64GB of RAM so a single job hasn't even been run to completion. But, it was able to launch and identify the process. Remember that BladeBit won't use a tmp directory and so tmp directory phase limits and such won't apply to BladeBit. Rather, you would generally set the global max jobs to 1. A global phase limit has been discussed but has not been implemented.

You rock. I've got a dual 7742 machine with 1TB of memory and some NVMe in a Raid-0 for the buffer (dst, as bladebit calls it). I'll let you know how it works out! Thank you, thank you. Are y'all taking donations yet? Or sponsorships? I'll gladly sponsor a project with a development team / author as responsive as you are.

altendky:
Thanks for the offer, so far I have declined donations and encouraged giving them to a charity instead.  https://github.com/ericaltendorf/plotman/issues/827

Let me know how it goes.  Depending, I may need to get access to a BladeBit-ready machine if significantly more development is needed.  Hopefully there won't be a lot though.  `:]`

altendky:
@jonathanhatch, have you found time to try this out?

jonathanhatch:
> @jonathanhatch, have you found time to try this out?

I haven't gotten it to work. Can you send me your config? There isn't a lot of guidance ie Bladebit in the current config file. I'm a little slow. You might need to walk me through this.

altendky:
Not a lot of config at this point, but sure, even then it would be good to document it.  Here's some of my config related to the plotters.  If it is trying to launch plotting processes you can check the plot logs.  The path can be configured at the top of the yaml.

```yaml
Reference#create
plotting:
    farmer_pk: b0a374845f4f4d6eab62fc4c5e17965d82ad7eee105818e5bd0cfcb46275a16acc4cd30955779bec841a716473416b21
    pool_contract_address: xch1u8ll2ztwhseej45d6u2zp9j4mlnzhwseccr0axqws9fl2tyj5u0svdy04y

    type: bladebit
    chia:
        k: 25
        e: False              # Use -e plotting option
        n_threads: 2         # Threads per job
        n_buckets: 128       # Number of buckets to split data into
        job_buffer: 6000     # Per job memory
    madmax:
        executable: /home/altendky/repos/chia-plotter/build/chia_plot
        n_threads: 8
    bladebit:
        executable: /home/altendky/repos/bladebit/.bin/release/bladebit
        threads: 2
```

jonathanhatch:
Brilliant. I'll give this a shot when my current plots have gone through.

altendky:
Regarding the topic of donations, FYI https://github.com/ericaltendorf/plotman/issues/827#issuecomment-922394860.

Issue #919, opened 2021-09-11
finding unused dst dirs fails because of trailing slash "/" 


FlyWieEinAirplane:
https://github.com/ericaltendorf/plotman/blob/77f85e3b38abc45d8cb6af3bad650b458c3aef35/src/plotman/manager.py#L130

var d will be in the format of "/mnt/destDir"
while each key in dir2ph.keys() will be in format "/mnt/destDir/"
Because of this, all directories will be in the ununsed_dirs array.

altendky:
I really need to get to using pathlib someday...  it wouldn't have this issue.  But, thank you for catching this and reporting it with an explanation.  I'll do a quick patchup for now.  You were observing this in the running of plotman I would guess?  With madMAx?

Any chance you could test out https://github.com/ericaltendorf/plotman/pull/920?

PR #920, opened 2021-09-12
.rstrip("/") for some dst comparisons

altendky:
long term just use pathlib.Path...

https://github.com/ericaltendorf/plotman/issues/919


FlyWieEinAirplane:
Unfortunately my drives are all full for now so I don' really have the space to test the scheduling. But I implemented a similar quick fix [here](https://github.com/FlyWieEinAirplane/plotman/commit/9a0585f3b00f0f2cc771b81e1d60b667a553adbb) which I observed to work just fine from which I suspect your change will also work.

PR #922, opened 2021-09-13
Release/v0.5.2

altendky:


altendky:
draft announcement

---

Hi @channel, it has been a bit quiet around here lately but it is now time for v0.5.2.  The headline change is that BladeBit support has been added.  If you can't make use of that due to having less than 416GB of RAM?  Well, join the club, neither can I.  `:]`  None the less, we try to keep your options open.  There are also a few other additions and fixes you can take a look at in the changelog at https://github.com/ericaltendorf/plotman/blob/v0.5.2/CHANGELOG.md.

ericaltendorf (approved):
Just did a cursory glance over changes.

PR #936, opened 2021-10-06
provide an explicit error for non-dict top level configuration file objects

altendky:


pieterhelsen:
Thanks for the fix! Maybe explicitly mention the possible cause?

```python
 if not isinstance(config_objects, dict):
        raise Exception(
            "plotman requires the top level configuration file object type be dict, got:"
            " {type(config_objects)}"
            "this could be due to a missing or corrupt config file."
        )
```

src/plotman/configuration.py
```diff
@@ -55,6 +55,12 @@ def get_validated_configs(
     schema = desert.schema(PlotmanConfig)
     config_objects = yaml.load(config_text, Loader=yaml.SafeLoader)
 
+    if not isinstance(config_objects, dict):
+        raise Exception(
+            "plotman requires the top level configuration file object type be dict, got:"
+            " {type(config_objects)}"
```
altendky:
```suggestion
            f"plotman requires the top level configuration file object type be dict, got:"
            f" {type(config_objects)}"
            f"\n    This could be due to an empty or corrupt config file.
```

src/plotman/configuration.py
```diff
@@ -55,6 +55,13 @@ def get_validated_configs(
     schema = desert.schema(PlotmanConfig)
     config_objects = yaml.load(config_text, Loader=yaml.SafeLoader)
 
+    if not isinstance(config_objects, dict):
+        raise Exception(
+            f"plotman requires the top level configuration file object type be dict, got:"
+            f" {type(config_objects)}"
+            f"\n    This could be due to an empty or corrupt config file.
```
altendky:
```suggestion
            f"\n    This could be due to an empty or corrupt config file."
```

src/plotman/configuration.py
```diff
@@ -55,6 +55,13 @@ def get_validated_configs(
     schema = desert.schema(PlotmanConfig)
     config_objects = yaml.load(config_text, Loader=yaml.SafeLoader)
 
+    if not isinstance(config_objects, dict):
+        raise Exception(
+            f"plotman requires the top level configuration file object type be dict, got:"
+            f" {type(config_objects)}"
```
altendky:
```suggestion
            f" {type(config_objects).__name__}"
```

altendky:
Thanks for the suggestion, included.

PR #937, opened 2021-10-13
Enhance Madmax arguments to support plotting for Chives

guydavis:
Adds the `-x` (network port) and `-k` (plot size) options to Madmax plotter configuration.  Also includes the boolean toggles of `-w` (wait for copy) and `-G` (tmp toggle).  Tested this in order to create the smaller k29/k30/k31 size plots that the Chives blockchain fork uses.

Each of these new options is fully optional so their absence in existing plotman.yaml config files will simply continue to use the previous implicit defaults.  This means no impact to existing Plotman users.

### DIFF
--- a/src/plotman/analyzer.py
+++ b/src/plotman/analyzer.py
@@ -163,7 +163,9 @@ def analyze(
                     )
                 )
             elif len(values) == 1:
-                row.append(plot_util.human_format(values[0], 1))
+                row.append(values[0])  # Use exact time in seconds.
+                # Due to Madmax and Bladebit speed, don't round times in seconds to nearest thousands
+                #row.append(plot_util.human_format(values[0], 1))
             else:
                 row.append("N/A")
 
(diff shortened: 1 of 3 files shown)


altendky:
The basic holdup here is just general support of plots that are not valid on Chia mainnet.  I want to avoid people testing a bit with k25 or such, forgetting to set it back, then plotting a pile of useless plots.  https://github.com/ericaltendorf/plotman/pull/277#issuecomment-839384900

So, the question to answer to get this merged is 'how do we make sure the user doesn't accidentally do the wrong thing?'.

Side note, sell this as 'supporting smaller plots for debugging' not 'chives support'.  `:]`

Oh, and thanks!

guydavis:
Hi, sorry for the slow reply.  Been busy with lots of other work.  You too it looks like.  Kudos on the recent Chia release with the new plotter integration.  Keep up the good work!

Regarding how to protect users from plotting accidentally small plots, how about a guard on the `network_port` value.  By default it is 8444, so would refuse to plot any k < 32.  However, user sets `network_port` to 9699 for Chives, then would allow plotting of k between 29 and 31?

PR #954, opened 2022-03-26
madMAx k34 support

altendky:


src/plotman/plotters/madmax.py
```diff
@@ -649,3 +665,131 @@ def _cli_974d6e5f1440f68c48492122ca33828a98864dfc() -> None:
...
+    "-v",
+    "--buckets3",
+    help="Number of buckets for phase 3+4 (default = buckets)",
+    type=int,
+    default=256,
+)
+@click.option(
+    "-t",
+    "--tmpdir",
+    help="Temporary directory, needs ~220 GiB (default = $PWD)",
```
ericaltendorf, line 725: i assume that's for k32?

src/plotman/plotters/madmax.py
```diff
@@ -649,3 +665,131 @@ def _cli_974d6e5f1440f68c48492122ca33828a98864dfc() -> None:
...
+    type=click.Path(),
+    default=pathlib.Path("."),
+    show_default=True,
+)
+@click.option(
+    "-2",
+    "--tmpdir2",
+    help="Temporary directory 2, needs ~110 GiB [RAM] (default = <tmpdir>)",
```
ericaltendorf, line 733: here too?

src/plotman/plotters/madmax.py
```diff
@@ -649,3 +665,131 @@ def _cli_974d6e5f1440f68c48492122ca33828a98864dfc() -> None:
...
+    default=False,
+)
+@click.option(
+    "-K",
+    "--rmulti2",
+    help="Thread multiplier for P2 (default = 1)",
+    type=int,
+    default=1,
+)
+def _cli_ecec17d25cd547fa4bb64b2eb7455b831c8a2882() -> None:
```
ericaltendorf, line 794: What are these?

src/plotman/plotters/madmax.py
altendky: Probably.  At the top of this list of options is a link to the code from which this was translated.  We don't expose this anywhere, I just decided to try to be complete so that this code is easier to think about.

src/plotman/plotters/madmax.py
altendky: Makes sense.

src/plotman/plotters/madmax.py
altendky: The `ecec17d...` is the commit hash of madmax from which this was translated.  It matches the link at the top of the command definition.  Presently we just use the latest one according to the version provided in the code like `@commands.register(version=(3,))`.  In the future we could tie some version detection code to make it pick the 'right' command definition for that version.

ericaltendorf (approved):

PR #968, opened 2023-02-01
Check free space available in dst before commencing plot

graemes:
A small tweak to check that there is sufficient space free in the dst folder to receive a new plot.  

This prevents the loop where a new plot is started only to fail because of a lack of space for the output and then repeats endlessly wasting power etc.

### DIFF
--- a/src/plotman/manager.py
+++ b/src/plotman/manager.py
@@ -183,8 +183,19 @@ def maybe_start_new_plot(
 
                     dstdir = max(dir2ph, key=key)
             
+            required_plot_space = plot_util.get_plotsize(32) + 10737418240
+            # If tmp & dst share directory and 'worst case' tmp space required
+            if plotting_cfg.type == "bladebit":
+                if dstdir == tmpdir:
+                    required_plot_space = required_plot_space + 64424509440  # 480 GiB
+            elif plotting_cfg.type == "madmax":
+                if dstdir == tmpdir:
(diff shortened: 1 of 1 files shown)


altendky (changes requested):
As I recall, the basic problem is that by the time the plot is done a transfer may have completed and freed up space.  Maybe this needs to be optional?  Also, `dst` is `tmp` for me so if we're checking space we need enough space for both things to avoid the general case of failing a plot for being out of space.

src/plotman/manager.py
```diff
@@ -183,105 +183,108 @@ def key(key: str) -> job.Phase:
...
                     dstdir = max(dir2ph, key=key)
 
-            log_file_path = log_cfg.create_plot_log_path(time=pendulum.now())
-
-            plot_args: typing.List[str]
-            if plotting_cfg.type == "bladebit":
-                if plotting_cfg.bladebit is None:
-                    raise Exception(
-                        "bladebit plotter selected but not configured, report this as a plotman bug",
+            if plot_util.df_b(dstdir) > plot_util.get_plotsize(32):
```
altendky: What if we invert this check and just return in that case and then avoid indenting everything else?

graemes: Makes sense.  Have also added a check for tmp/dst and space required.

graemes: re: your first comment - the newly freed space will be detected on a subsequent check (after the appropriate refresh time)?

